PR #23, opened 2015-01-05
Mint inherits from requests.Session.

brycedrennan:
Just a little cleaner and allows for future optimizations.


mrooney:
I didn't know this was possible, thanks!

PR #24, opened 2015-01-05
Cleanup

brycedrennan:
Functionally behaves the same, but simplified lots of code and made code conform with PEP8 style guide.

Will build additional features on top of this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@ dist
 __pycache__
 *.pyc
 *.egg-info
-
+/.idea
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -2,7 +2,6 @@ import datetime
 import json
 import random
 import requests
-import ssl
 import time
 import xmltodict
 
@@ -21,7 +20,6 @@ class MintHTTPSAdapter(HTTPAdapter):
         self.poolmanager = PoolManager(num_pools=connections, maxsize=maxsize, **kwargs)
 
 class Mint(requests.Session):
-    json_headers = {"accept": "application/json"}
     request_id = 42 # magic number? random number?
     token = None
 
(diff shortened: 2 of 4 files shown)


mcronce:
I don't really see a ton of value in changing code _just_ to get it to conform to PEP8.  In particular, replacing tabs with spaces actually _loses_ editor functionality.

brycedrennan:
The principal value is that it's easier for other coders to read and work with since it's what they're used to seeing.  

Most editors I've worked with don't care about tabs vs spaces.  I highly recommend PyCharm if you're looking for a new editor.  

It's your project so if you think it's not useful that's okay with me.

mcronce:
I wouldn't call it my project, just a contributor ;)

That said, many people are using a terminal and as a result, vim/emacs.  The advantage tabs provide is that the viewer defines the width - if I like 4-character indentation and you like 32-character indentation, we can both have that.  That crusade Guido has against tabs is a little ridiculous if you ask me.

The difference in readability (of the whole thing, not of the tabs->spaces conversion, there's no realistic change there) seems pretty minimal to me but again, I wouldn't call it my project ;)

brycedrennan:
Okay, well I don't care that much. I didn't even know I had done tabs->spaces conversion to be honest.  PyCharm must just do that automatically. Personally I don't care about tabs vs spaces but I do like to conform to standards, especially since all the code quality tools just work if I conform.

I am probably not going to be motivated enough to manually remove the PEP8 changes from my other changes. I bet there is a way to undo that specific commit. Maybe I'll look that up later.

mcronce:
git revert could do it but depending on what changes were committed
afterward could make irritating conflicts you'd need to resolve.  Like I
said it's not my project either (although I do feel strongly ;)) but the
space-tab conversion at least is easy enough to do in sed or a vim macro

PR #28, opened 2015-01-18
get transactions as pandas dataframe

adamgreenhall:


mrooney:
Awesome, thanks! Is this something that might make sense from the command line as well? Perhaps either dumping the CSV to stdout, or serializing to JSON? If so it would be awesome to add that to the main() method with the other command-line options and update the README.

mrooney:
Let me know your thoughts @adamgreenhall , I'd love to get this merged as it has been requested before :)

adamgreenhall:
@mrooney - can do. will work on that this morning.

adamgreenhall:
@mrooney - think this is ready now. review?

mrooney:
Great thanks, I'll get this merged @adamgreenhall :) Does the setup.py entry_points mean we don't need the bin/mintapi anymore? I wasn't familiar with this feature but it looks neat.

adamgreenhall:
yup - with the entry_points added, after you install, you can call mintapi directly from the bash shell.

mrooney:
Perfect, thanks, I removed the bin/ directory which we don't need now. There were two small bugs I fixed that you may want to review at https://github.com/mrooney/mintapi/commit/432fc2ed0b83ed62c3a7873e9b8ed7eac584afa8. Particularly, if no filename was specified with transactions, it crashed because the second wasn't an elif and it tried to call endswith() on None. And second, if you didn't specify a csv or json extension, it silently did nothing. Do these make sense? Thanks again for your awesome contribution!

mrooney:
And, this is merged and released as 1.6 :D

adamgreenhall:
looks good - thanks!

PR #30, opened 2015-02-08
use keyring to handle password storage

adamgreenhall:
@mrooney - also wanted to use keyring to handle password storage. 

example flow:

    python mintapi/api.py -u someone@gmail.com
    Password: ****
    [ {json_response: ...}]
   
    # now next time, password for this user gets looked up from the OS keyring
    python mintapi/api.py -u someone@gmail.com
    [ {json_response: ...}]


mrooney:
Really cool, thanks for another great feature! Merged and released as 1.7. The only change I made was to still allow passing credentials on the command-line, even though keyring is better, simply because someone using the old method and upgrading via pip could run into a headache trying to figure out why their flow is broken.

PR #32, opened 2015-02-13
Fixing get_transactions() call using old session syntax

zabracks:
The get_transactions() call was using the self.session.get(...) syntax, but this doesn't work now that the class inherits from requests.Session.

Fixed and tested locally, but didn't write tests, only because test coverage is fairly low on this project anyway.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -17,11 +17,15 @@ try:
 except ImportError:
     pd = None
 
-DATE_FIELDS = [
+ACCOUNT_DATE_TIMESTAMP_FIELDS = [
     'addAccountDate',
     'closeDate',
     'fiLastUpdated',
     'lastUpdated',
+]
+
+ACCOUNT_DATE_STRING_FIELDS = [
+    'dueDate'
 ]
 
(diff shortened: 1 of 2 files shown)


mrooney:
Great catch, thank you! I'm not sure I follow all the changes, any particular reason why https://github.com/mrooney/mintapi/commit/4faeb7dc1e6f1d2d1a8d69cef0c5134051b3cef8 wouldn't be all that is necessary? It seems to work on my end, apologies if I'm totally misunderstanding what you are trying to accomplish :)

ianlevesque:
For what it's worth it was broken for me too without this change.

ianlevesque:
The single linked commit I mean.

mrooney:
Great, should be fixed and released as 1.8! If there are other cool features your pull request address @zabracks , please do open a pull request explaining them and I'll happily review them!

zabracks:
Sorry, made the pull request, then kept on developing without thinking! 

I found a different method to get transactions from a JSON blob rather than the csv. The majority of the other commits deal with that.



Drew Worthey

PR #34, opened 2015-03-20
Added pandas as a requirement

advancemh:
Since it's needed to for the transactions


mark-adams:
If everything else works without pandas, it seems like it would make more sense to leave it the way it is so people aren't forced to install such a heavy library just to look at budget or account data.

mrooney:
Yes, I think it is fine to just fail for commands that need pandas until pandas is installed, since it does have a helpful error message. The requirements can just be for packages required to do the base operation of getting accounts and balances, or those packages which are small and always available.

PR #37, opened 2015-04-19
Modified usage of `keyring` to avoid an exception on platforms that don't support the keyring library.

### LINKED ISSUE #35: import of keyring should be lazy
Keyring doesn't work on some platforms (notably AppEngine).  This could be fixed by lazy loading it in main() but it is still listed as a required package in setup.py.  Should it still be a strict requirement?

mark-adams:
Fixes #35

Also:

- Removed `keyring` from requirements since it now handles the missing library gracefully
- Changed from using the older `optparse` to the newer `argparse` library
- Replaced `--user` flag with `--keyring` and better optional positional username and password argumetns
- Updated the README

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -4,7 +4,6 @@ import random
 import requests
 import time
 import xmltodict
-import keyring
 
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.poolmanager import PoolManager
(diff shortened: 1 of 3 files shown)


mrooney:
I dig it, thanks! The only issue is that this isn't backwards compatible with the use case of just passing a username; anyone doing that now will be broken and an upgrade and have to magically know to add the --keyring argument. I'd probably make line 346 something like:

    password = (keyring and keyring.get_password('mintapi', email)) or getpass.getpass("Mint password: ")

This way, if there isn't a password but there is one in the keyring, that'd be used, and I can't imagine why we should ignore a keyring password if one exists. What do you think?

mintapi/api.py
```diff
@@ -331,38 +333,68 @@ def initiate_account_refresh(email, password):
...
+    if options.keyring and not keyring:
+        cmdline.error('--keyring can only be used if the `keyring`'
+                      'library is installed.')
 
-    # Handle Python 3's raw_input change.
     try:
-        input = raw_input
+        from __builtin__ import raw_input as input
     except NameError:
         pass
```
mark-adams: This code for fixing raw_input didn't really work. It caused an error for me on Python 3.
Apparently in Python 3, even trying to assign to input even if you fail, causes the built-in input to be blasted away.

The new code I added works in both Python 2 and Python 3 and is based on how the `six` compatibility library handles things.

mintapi/api.py
```diff
@@ -331,38 +333,68 @@ def initiate_account_refresh(email, password):
...
+    password = options.password
+
+    if not email:
+        # If the user did not provide an e-mail, prompt for it
+        email = input("Mint e-mail: ")
+
+    if keyring and not password:
+        # If the keyring module is installed and we don't yet have
+        # a password, try prompting for it
         password = keyring.get_password('mintapi', email)
```
mark-adams: I think this does what you're looking for now.

Basically, specify `--keyring` (369) but its not installed, you get an error. However, if you don't specify `--keyring` and it IS installed, this code will try and retrieve your password from the keyring if the code hasn't already discovered a password.

PR #38, opened 2015-04-19
Reformatted code to follow standard PEP8 conventions

mark-adams:
The Python standard library and a lot of other open source libraries choose to follow the [PEP8](https://www.python.org/dev/peps/pep-0008/) style guide for Python code to help make sure code style is consistent and as readable as possible.

This PR takes the existing code and makes numerous improvements to help it match up with PEP8.
There are no changes at all that affect the way the code functions. All the changes are purely formatting.

* One exception: I removed the `import ssl` since it is not being used anywhere in the code.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -2,7 +2,6 @@ import datetime
 import json
 import random
 import requests
-import ssl
 import time
 import xmltodict
 import keyring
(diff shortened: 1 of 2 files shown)


mrooney:
Thank you for your time on this! Is there an automated command for this? I wonder if it might be worth adding to the Travis build file so we can catch PEP-8 issues going forward.

mark-adams:
Its funny you should mention that. I created a branch for that exact thing the other day but wanted to wait until the code was all fixed up with regards to PEP8 before suggesting a merge.

I've introduced a PR; its now #39

PR #41, opened 2015-05-22
Create requirements.txt based on setup.py dependencies

mplewis:
This makes it easier for other developers to start working on the project. Now they can just `pip install requirements.txt` and get going.

I left out the optional requirements, like pandas and pypandoc.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+xmltodict==0.9.2
+requests==2.7.0


mark-adams:
requirements.txt makes a lot of sense for apps, but not for libraries.

The exact same thing can be accomplished as is by doing `pip install -e .`

mrooney:
I don't have any strong feelings on this really. a `pip install mintapi` should pull in the required ones, `requests` and `xmltodict`, correct?

PR #42, opened 2015-05-22
Fix StringIO issues in Python 3

mplewis:
The `StringIO` module was moved from the `StringIO` package in Python 2 to the `io` package in Python 3. Lines 150-155 check try the Python 2 import first, then fall back to Python 3, [as prescribed here](http://python3porting.com/stdlib.html).

`result.content` is a `str` object in Python 2, but a `bytes` (Unicode) object in Python 3. We try just shoving it into a StringIO first (the Python 2 way), then fall back to converting the bytes to a string if that fails (the Python 3 way).

Nosetests pass on Python 2 and 3. Getting transactions passes in both 2 and 3 as well, tested with this script:

```
import mintapi
mint = mintapi.Mint('matt@mplewis.com', 'chocolate-chip-cookie-dough')
txns = mint.get_transactions()
```

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -2,6 +2,11 @@ import json
 import random
 import time
 
+try:
+    from StringIO import StringIO  # Python 2
+except ImportError:
+    from io import BytesIO as StringIO  # Python 3
+
 from datetime import date, datetime, timedelta
 
 import requests
@@ -148,11 +153,6 @@ class Mint(requests.Session):
         if not pd:
             raise ImportError('transactions data requires pandas')
 
-        try:
-            from StringIO import StringIO  # Python 2
-        except ImportError:
-            from io import StringIO  # Python 3
-
         result = self.get(
             'https://wwws.mint.com/transactionDownload.event',
             headers=self.headers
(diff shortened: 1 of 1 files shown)


mintapi/api.py
```diff
@@ -147,7 +147,12 @@ def get_accounts(self, get_detail=False):  # {{{
     def get_transactions(self):
         if not pd:
             raise ImportError('transactions data requires pandas')
-        from StringIO import StringIO
+
+        try:
+            from StringIO import StringIO  # Python 2
+        except ImportError:
+            from io import StringIO  # Python 3
```
mark-adams: Why not change this to:
```
except ImportError:
    from io import BytesIO as StringIO
```

Then you could change 167 - 169 to simply be:
```
s = StringIO(result.content)
```
and it would work all the time.

mrooney:
Thanks mplewis for your time and contribution! Mark's suggestions sound good to me. Is this something using the six library might solve as well? It might be worth pulling it in based on other issues that have come up, but I don't have any strong opinions.

mintapi/api.py
mplewis: Good idea! I'll make those changes.

mplewis:
I thought about that too. The six library might solve this kind of thing well. I would think about using six if more py2/3 issues popped up besides this one.

mrooney:
Thanks mplewis!

PR #43, opened 2015-07-14
Factor out get_transactions_csv API method for retrieving raw CSV data

jbms:
This simply separates the logic for retrieving the CSV transactions data from
the logic for parsing it as a Pandas DataFrame.  This allows users to retrieve
the raw transaction data without depending on Pandas.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -149,6 +149,24 @@ class Mint(requests.Session):
             accounts = self.populate_extended_account_detail(accounts)
         return accounts
 
+    def set_user_property(self, name, value):
+        url = 'https://wwws.mint.com/bundledServiceController.xevent?legacy=false&token=' + self.token
+        req_id = str(self.request_id)
+        self.request_id += 1
+        result = self.post(
+            url,
+            data = {'input': json.dumps([{'args': {'propertyName': name,
+                                                   'propertyValue': value},
+                                          'service': 'MintUserService',
+                                          'task': 'setUserProperty',
+                                          'id': req_id}])},
+            headers=self.json_headers)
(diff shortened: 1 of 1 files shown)


jbms:
I also added support for retrieving investment transactions, and for retrieving JSON transaction data, which includes some additional information not present in the CSV data.

mintapi/api.py
```diff
@@ -149,19 +149,105 @@ def get_accounts(self, get_detail=False):  # {{{
...
+                    offset=offset,
+                    rnd=Mint.get_rnd(),
+                    query_options=(
+                        'accountId=0&task=transactions' if include_investment
+                        else 'task=transactions,txnfilters&filterType=cash'))
+            result = self.get(url)
+            if result.status_code != 200:
+                raise ValueError(result.status_code)
+            if not result.headers['content-type'].startswith('text/json'):
+                raise ValueError('non-json content returned')
```
mrooney: There's now some duplicated code where we do a self.get, followed by checking the status and code content-type. Let's factor this into a get_and_check method that takes a content type as an argument, and does the get, status code, and content-type check!

mrooney:
Would you mind merging master and pushing, as the check should now be resolved on master? And once that little chunk of duplicated code is gone, I'll merge this and everyone can enjoy your fine work!

jbms:
Done.  Pretty much all of the direct calls to self.post and self.get (that don't currently check the status at all) could be replaced by request_and_check, but I figured I'd leave such a change up to you.

mrooney:
Great, thank you, merged and released to pypi as 1.12 :)

Issue #16, opened 2014-10-20
Net Worth?


elijahchancey:
How hard would it be to add a function that grabs your net worth (All Assets, including investments minus All Debts)? Ps Thanks for your hard work!

mplewis:
I noticed the same thing as @kunalbhat. Here are the account types I have in my Mint account and their "value types".

**Regular**: positive value = assets, negative value = debt
**Inverse**: positive value = debt, negative value = assets

Account type | Value type
--------- | ----------------------
`credit` | Regular
`loan` | Inverse
`bank` | Regular
`investment` | ??
`vehicle` | Regular

elijahchancey:
This is a great help. Thanks!

PR #44, opened 2015-08-10
Net worth

wendlinga:
First attempt at fixing issue #16 . Should add the net worth capability and the numbers look correct with my mint accounts values. Let me know if they're not correct for your account and i can fix it.


mrooney:
Awesome, thanks, I merged this, added a command line option and a few other tweaks, and released it as 1.13. Thanks for your contribution!

PR #46, opened 2015-10-07
Build budget categories by ID to handle top-level categories

tylerneu:
Here is how I handled my own budget utilizing top-level categories.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -296,24 +296,23 @@ class Mint(requests.Session):
                 budget['cat'] = self.get_category_from_id(budget['cat'])
 
         return budgets
-    
+
     def get_category_from_id(self, cid):
         if cid == 0:
             return 'Uncategorized'
-         
+
         categories = self.get_categories()
 
         for i in categories:
             if categories[i]['id'] == cid:
-               return categories[i]['name']
-               
+                return categories[i]['name']
+
             if 'children' in categories[i]:
                 for j in categories[i]['children']:
                     if categories[i][j]['id'] == cid:
                         return categories[i][j]['name']
 
         return 'Unknown'
-    
 
     def initiate_account_refresh(self):
         # Submit refresh request.
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -151,7 +151,10 @@ class Mint(requests.Session):
 
     def get_transactions(self):
         if not pd:
-            raise ImportError('transactions data requires pandas; please pip install pandas')
+            raise ImportError(
+                'transactions data requires pandas; '
+                'please pip install pandas'
+            )
         result = self.get(
             'https://wwws.mint.com/transactionDownload.event',
             headers=self.headers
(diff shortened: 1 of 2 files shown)


mintapi/api.py
```diff
@@ -151,7 +151,10 @@ def get_accounts(self, get_detail=False):  # {{{
 
     def get_transactions(self):
         if not pd:
-            raise ImportError('transactions data requires pandas; please pip install pandas')
+            raise ImportError(
+                'transactions data requires pandas; '
+                'please pip install pandas'
+            )
```
mrooney: It would be great to avoid unrelated changes. Is this not PEP 8 already?

setup.py
```diff
@@ -6,7 +6,9 @@
     from pypandoc import convert
     read_md = lambda f: convert(f, 'rst')
 except ImportError:
-    print("warning: pypandoc module not found, could not convert Markdown to RST")
+    print(
+        "warning: pypandoc module not found, could not convert Markdown to RST"
+    )
```
mrooney: Similarly, let's avoid this unless there is a compelling reason.

mintapi/api.py
```diff
@@ -263,8 +266,6 @@ def get_categories(self):  # {{{
         return categories
 
     def get_budgets(self):  # {{{
-        # Get categories
-        categories = self.get_categories()
```
mrooney: I think it would be best to keep this here, so we can pass it to get_category_from_id, otherwise it is calling self.get_categories() quite a few times for no particular reason. Even if it happens to be a no-op after the first call, that's still an implementation detail that we shouldn't assume :)

mintapi/api.py
```diff
@@ -293,10 +294,27 @@ def get_budgets(self):  # {{{
         # Fill in the return structure
         for direction in budgets.keys():
             for budget in budgets[direction]:
-                budget['cat'] = categories[budget['cat']]
+                budget['cat'] = self.get_category_from_id(budget['cat'])
 
         return budgets
 
+    def get_category_from_id(self, cid):
```
mrooney: Take categories as an argument here to avoid self.get_categories().

mrooney:
Thanks so much for your time Tyler! If you could just address the few comments, I think this would be ready to go :)

mintapi/api.py
tylerneu: I agree but [it wasn't PEP 8.](https://travis-ci.org/mrooney/mintapi/jobs/84168110)

setup.py
tylerneu: Same as above. I only made this change to satisfy my own execution of `flake8` but in retrospect Travis CI never complained. I'll back this out.

mintapi/api.py
tylerneu: I understand and agree with your suggestion.

tylerneu:
I have finished my changes based on your comments. Let me know if there is anything else I can do.

mintapi/api.py
mrooney: Apologies, thanks!

PR #47, opened 2015-11-01
Get JSON transactions as a Generator

bfreis:
This is a very simple change that includes a method, `get_transactions_json_generator`. This method is almost exactly the same as the original `get_transactions_json`, but returns a generator instead.

I also refactored the `get_transactions_json` method to use this new generator, and return a `list(...)` of it, preserving the exact functionality.

This gave me a nice performance boost for many tasks that require just the very recent history of my transactions!

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -210,15 +210,18 @@ class Mint(requests.Session):
         interface.  Note that the CSV transactions never exclude duplicates.
 
         """
-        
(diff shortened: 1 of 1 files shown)


mrooney:
Looks like this got lost, my apologies. If you are willing to update to resolve conflicts, I'm happy to review and merge!

PR #48, opened 2015-11-09
Fix: flake8 errors

yegle:
Fixed two flake8 errors that caused the TravisCI's build error.


bfreis:
Ha, looks like this is why my PR wasn't building! Now it should go smooth :)

yegle:
Ping? The `0L` fix is critical because Python3 removed the `L` suffix to numbers.

mrooney:
Thanks!

PR #50, opened 2015-11-21
Beef up json pull and create new detailed pull

dtiz:
This update allows the json pull to be limited to a start date (in format mm/dd/yy). This speeds up the pull time dramatically if you have lots of transactions.

Also created a new beefier pull that starts with the json pull, then converts the dates, reverses the credits, and drops pending transactions (because pending transactions sometimes change when they actually post). Added two helper functions for the new one, so maybe those can be pulled in as lambdas.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -246,8 +246,9 @@ class Mint(requests.Session):
             if start_date:
                 dates = list(df['odate'])
                 try:
-                    first_date = datetime.strptime(dates[0] + 
-                        str(datetime.isocalendar(date.today())[0]), '%b %d%Y')
+                    first_date = datetime.strptime(dates[0] +
+                                 str(datetime.isocalendar(date.today())[0]),
+                                 '%b %d%Y')
                 except:
                     first_date = datetime.strptime(dates[0], '%m/%d/%y')
                 if first_date < start_date:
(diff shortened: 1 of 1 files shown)


mintapi/api.py
```diff
@@ -237,13 +242,73 @@ def get_transactions_json(self, include_investment=False,
...
+        format, and reverses credit activity.
+
+        Note: start_date must be in format mm/dd/yy. If pulls take too long,
+        use a more recent start date. See json explanations of
+        include_investment and skip_duplicates.
+
+        Also note: Mint includes pending transactions, however these sometimes
+        change dates/amounts after the transactions post. They have been
+        removed by default in this pull, but can be included by changing
+        include_pending to False
```
mrooney: Should this be "remove_pending" instead of include_pending?

mintapi/api.py
```diff
@@ -197,24 +197,29 @@ def set_user_property(self, name, value):
         if req_id not in response:
             raise Exception("Could not parse response to set_user_property")
 
-    def get_transactions_json(self, include_investment=False,
+    def get_transactions_json(self, start_date=None, include_investment=False,
```
mrooney: Let's make start_date the last argument, in case anyone is using this API elsewhere with positional arguments, so we don't break their code.

mintapi/api.py
```diff
@@ -237,13 +242,73 @@ def get_transactions_json(self, include_investment=False,
...
+        change dates/amounts after the transactions post. They have been
+        removed by default in this pull, but can be included by changing
+        include_pending to False
+
+        """
+        if not pd:
+            raise ImportError(
+                'transactions data requires pandas; '
+                'please pip install pandas'
+            )
```
mrooney: This chunk of code is now duplicated. Would you mind making a function like "assert_pd" that has this code and then can just be called in one line in the places where we use it?

mrooney:
Looks great, thanks so much for your time and contribution, I'll merge and release this as 1.16 now :)

PR #52, opened 2015-12-08
Update api.py

dancudds:
PoolManager has moved from requests and exists in urllib3


dancudds:
Fix for missing poolmanager in requests

Issue #57, opened 2015-12-18
JSON pull issue


dtiz:
For some reason the JSON pulls are throwing an error with the request_and_check. Not sure why this happens now. It's saying the content type 'application/son;charset=UTF-8' does not match 'text/json'

Going to that site, the data looks fine and seems to be in JSON format.

tylerneu:
I'm assuming the returned Content-Type changed after `get_transactions_json` was added. I have submitted a pull request that simply updates the expected header value.

pca2:
+1

mrooney:
Looks great, thanks, merged and released as 1.17!

PR #59, opened 2015-12-20
replaced text/json with application/json

scsinutz:
This pull request is to address the error: 'application/son;charset=UTF-8' does not match 'text/json' when calling mint.get_transactions_json()


mrooney:
Thanks, this was also #57 and has been fixed and released as 1.17!

PR #62, opened 2015-12-30
Add requirements.txt

andersonvom:
This adds the standard requirements file for python projects, which
allows users and developers to easily install all required packages in
order to make the app work and move them away from the setup.py file.
This makes it easier, for example, to tell pip where to get the packages
from, among other things.

Usage: pip install -r requirements.txt

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+requests
+xmltodict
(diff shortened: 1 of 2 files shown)


mrooney:
Thanks, though we need a setup.py for pypi and I'm not a fan of duplicating requirements. We can just use "pip install -e ." already. Though, I'd happily take a PR to improve the documentation for this! See https://github.com/mrooney/mintapi/pull/41 for more details.

PR #63, opened 2016-01-11
Match content types more robustly, and eliminate unnecessary pandas dependency

jbms:


mintapi/api.py
```diff
@@ -265,18 +266,15 @@ def get_transactions_json(self, include_investment=False,
...
-                last_dt = self._dateconvert(dates[-1])
+                last_dt = self._dateconvert(txns[-1]['odate'])
                 if last_dt < start_date:
-                    keep_txns = []
-                    for item in txns:
-                        if self._dateconvert(item['odate']) >= start_date:
-                            keep_txns.append(item)
+                    keep_txns = [
+                        item for item in txns
+                        if self._dateconvert(item['odate']) >= start_date]
```
mrooney: Why do we only consider the last item in this case, whereas before we were iterating over all of them?

mintapi/api.py
```diff
@@ -1,6 +1,7 @@
 import json
 import random
 import time
+import re
```
mrooney: Do you mind making this alphabetical and also removing the pandas import logic and the pandas mention the README?

mintapi/api.py
mrooney: Sorry, I don't follow. If you did "for item in txns:" you'd be looping over them. But line 276 is just "item for item in txns" which is just a statement. The next line isn't inside any looping scope, so the last item is going to be the value of item and that will be executed just once. Right? Am I missing something?

jbms: This is a Python list comprehension expression.  You can read about them here: http://www.secnetix.de/olli/Python/list_comprehensions.hawk

It is equivalent to:
keep_txns = []
for item in txns:
  if self._dateconvert(item['odate']) >= start_date:
    keep_txns.append(txn)

mrooney: Oh, yes I'm very familiar with them, I just didn't see the square brackets above so I thought 276 was its own line. Now I understand what is happening syntactically, thanks! :)

mrooney:
Thanks! Merged, tested, and released to pypi as 1.22 :)

PR #77, opened 2016-06-03
Fix #73: Login Doesn't Work

### LINKED ISSUE #73: Login Doesn't Work...
running `mint.get_accounts()` from python, I get the following error:

```
Exception('Could not parse account data: ' + response)
Exception: Could not parse account data: <error><code>1</code><description>Session has expired.</description><name></name><type></type></error>
```

running from the command line, I get this:

```
Exception: Mint.com login failed[1]
```

hiromu2000:
This fixes the login issue #73 .
Please note that this patch changes the api, i.e., the input variables for the constructor.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -52,11 +52,11 @@ class Mint(requests.Session):
     request_id = 42  # magic number? random number?
     token = None
 
-    def __init__(self, email=None, password=None):
+    def __init__(self, email=None, password=None, ius_session=None):
         requests.Session.__init__(self)
         self.mount('https://', MintHTTPSAdapter())
         if email and password:
-            self.login_and_get_token(email, password)
+            self.login_and_get_token(email, password, ius_session)
 
     @classmethod
     def create(cls, email, password):  # {{{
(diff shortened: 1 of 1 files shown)


mrooney:
Thanks! This isn't working for me, I notice that the response I get from POSTing to https://accounts.mint.com/access_client/sign_in looks like a CAPTCHA challenge:

    {"iamTicket":null,"action":"CHALLENGE","riskLevel":"MED","challenge":[{"primary":true,"type":"CAPTCHA","value":null,"country":null,"tokenFormat":null},{"primary":false,"type":"CARE","value":null,"country":null,"tokenFormat":null}]}

If I login in via the web, it works fine, just not from the command-line. Did you experience this issue?

hiromu2000:
Thanks for merging. I get the same CAPTCHA challenge unless I login with a web browser.
First, login with a browser and retrieve the ius_session from the browser. Then login with this python api with the retrieved ius_session.
Please remember that I have to do a browser login every a few days, otherwise I get the challenge.
You don't have to change the ius_session, but have to login normally when you get the issue.

PR #83, opened 2016-11-02
Automate (as far as is reasonable) extracting the ius_session.

matthewwardrop:
This solves issue #73, at least on non-server machines.

If you simply run:
```
import mintapi
m = mintapi.Mint(<username>, <password>)
```
A chrome webdriver pops up that logs you in, and then close once the ius_session has been extracted. If MFA is required, handle this manually, and then the window will close. This does require installation of the chromedriver; which is easily done in most Linux distributions and by `brew` on Mac OS X. No idea about Windows.

Note: Build is failing due to pep8 style errors pre-my patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,11 +3,23 @@ mintapi
 
 a screen-scraping API for Mint.com. [![Build Status](https://travis-ci.org/mrooney/mintapi.svg?branch=master)](https://travis-ci.org/mrooney/mintapi)
 
-Requirements
+Installation
 ===
 Ensure you have Python 2 or 3 and pip (`easy_install pip`) and then:
 
     pip install mintapi
+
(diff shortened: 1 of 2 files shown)


mrooney:
Thanks! This is a great improvement and I'm happy to merge it. Could you just provide some installation instructions in the README (and if you merge master, the pep8 issue is fixed). I'd do it myself but "brew search chromedriver" doesn't return anything useful so I'm not sure what to install :)

matthewwardrop:
@mrooney I've rebased (and slightly modified) my commits atop master. Let me know if you need anything else done :).

Note that I'm not sure what `thx_guid` is; so I only trigger the automatic grab if `ius_session` is not specified (which is consistent with the tests).

mrooney:
Thanks, so awesome! Merged, cleaned up the README a bit, and releasing as 1.20. Thanks for making this a lot easier to use again :)

matthewwardrop:
@mrooney No worries. Glad to be of use :).

PR #84, opened 2016-11-05
2-step work around

boondawgy:
This commit add two cookies to the login step.  Providing these cookies
will bypass the need to 2-step authenticate.  I've run this code daily
for two weeks now and it hasn't failed yet.  I don't think this is the
best solution; but it works and I know others can benefit from it.


hiromu2000:
+1. It works for me.

george-lamperis:
I can also confirm that this works for me. @mrooney do you plan to merge this?

PR #92, opened 2017-02-27
Add MintException exception class

titilambert:
This PR create a new Exception class, called MintException.
This is useful when you use mintapi as a librairy, in order to not catch all exceptions.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -46,6 +46,10 @@ MINT_ROOT_URL = 'https://mint.intuit.com'
 MINT_ACCOUNTS_URL = 'https://accounts.intuit.com'
 
 
+class MintException(Exception):
+    pass
+
+
 class MintHTTPSAdapter(HTTPAdapter):
     def init_poolmanager(self, connections, maxsize, **kwargs):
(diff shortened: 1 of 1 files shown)


mintapi/api.py
```diff
@@ -171,7 +175,7 @@ def get_session_cookies(self, username, password):
         try:
             from selenium import webdriver
             driver = webdriver.Chrome()
-        except Exception as e:
+        except MintException as e:
```
mrooney: I don't believe this should be MintException. Other python exceptions (ImportError among others) could be raised and we want to catch them all!

titilambert: Ho ! You're right !

titilambert:
Change done !

mrooney:
Thank you very much for your contribution!

titilambert:
Thank you !
Do you have an ETA for the next release ?
I would like to use the next release in the PR: https://github.com/home-assistant/home-assistant/pull/6132

mrooney:
Released today as 1.23! :)

titilambert:
Thanks !

Issue #66, opened 2016-01-17
get_budgets returns only the most recent month, not the history for the past year


ashleydb:
The json data returned from mint.com contains the full history of budget data, but the output from mintapi only shows the current month.

I'm not familiar with Python, but I suspect it is related to one of these sections of setting up and filling out the return data. I feel like there is a loop missing...

```
    # Make the skeleton return structure
    budgets = {
        'income': response['data']['income'][
            str(max(map(int, response['data']['income'].keys())))
        ]['bu'],
        'spend': response['data']['spending'][
            str(max(map(int, response['data']['income'].keys())))
        ]['bu']
    }

    # Fill in the return structure
    for direction in budgets.keys():
        for budget in budgets[direction]:
            budget['cat'] = self.get_category_from_id(
                budget['cat'],
                categories
            )
```

mrfiat:
Is the MintAPI working for you right now?  I cannot get it to get transactions.  See my bug report.  Thx.

ashleydb:
I didn't try get transactions but the default call, (accounts,) and budgets
worked for me.

On Sunday, 17 January 2016, mrfiat <notifications@github.com> wrote:

> Is the MintAPI working for you right now? I cannot get it to get
> transactions. See my bug report. Thx.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mrooney/mintapi/issues/66#issuecomment-172331767>.
>


-- 


Ashley Bennett
www.HiAsh.com

spearna:
I also wanted a budget history so I can trend my budget actuals in my local monthly budget tool.  
I looked into this and worked a solution.  Currently, the 'budgets' dictionary in the api is being created just with the max month number (ie most current). I implemented a 'hist' argument in 'get_budget()' that allows for a select-able range from 1-12 months. 
I have it ready to go, just need to get set up to create a Pull Request. Do I have to be approved somehow to do this?

PR #94, opened 2017-03-16
add history to get_budget

spearna:
I also wanted a budget history so I can trend my budget actuals in my local monthly budget tool.
I looked into this and worked a solution. 
Currently, the 'budgets' dictionary in the api is being created just with the max month number (ie most current). I implemented a 'hist' argument in 'get_budget()' that allows for a select-able range from 1-12 months.
I sling code but don't claim to be efficient, so pythonic edits are welcome.
Closes Issue #66

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -518,7 +518,7 @@ class Mint(requests.Session):
 
         return categories
 
-    def get_budgets(self):  # {{{
+    def get_budgets(self, hist=None):  # {{{
         # Issue request for budget utilization
         today = date.today()
         this_month = date(today.year, today.month, 1)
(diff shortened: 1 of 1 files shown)


mintapi/api.py
```diff
@@ -537,23 +533,67 @@ def get_budgets(self):  # {{{
...
-            'spend': response['data']['spending'][
-                str(max(map(int, response['data']['income'].keys())))
-            ]['bu']
-        }
+        if hist is not None:  # version proofing api
+            def mos_to_yrmo(mos_frm_zero):
+                date_yr_mo = datetime(year=int(mos_frm_zero / 12),
+                                      month=mos_frm_zero % 12 + 1,
+                                      day=1).strftime("%Y%m")
+                return date_yr_mo
```
mrooney: Might as well just directly return the created datetime instead of creating a variable and then returning it on the next line (ie, return datetime(...))

austincrft:
@mrooney does this require additional changes to get merged? I'd love to take advantage of this in my scripts.

spearna:
Thanks for pinging on this, @austincrft.  I believe I had made the requested edits, but I'd like a confirmation on that.

mrooney:
Thanks, seems reasonable! Just needs a merge with the master so it can cleanly merge and then I can merge and release :) Thanks again for your contribution and patience.

PR #99, opened 2017-05-28
Excel compatible dates

felciano:
This patch updates the “presentable” accounts format to use a YYYY-MM-DD HH:MM:SS
format that Excel will parse automatically. Note that the original code used `repr(v)` which typically rendered a python snippet of form datetime.datetime(2017, 05, 05, 12, 55, 00) which seems to be of
limited value in an export

(apologies if this looks convoluted in the commits list; this is my first public pull request so I it isn't as clean as I would like. The "files changed" shows the small code changes that are the essence of this pull request)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,6 @@ install:
     - python setup.py install
     - pip install nose flake8 pep8-naming
 before_script:
-    - flake8 mintapi
+    - flake8 --ignore=E501 mintapi # pep8 flake ignoring long lines
 # command to run tests
 script: nosetests
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+1.24
+---
+- fix "Unknown" categories when categories in budgets (thanks @drunnells!)
+
+1.23
+---
+- mintapi raises MintException instead of Exception for better exception
+  handling when used as a library (thanks @titilambert!)
+
 1.22
 ---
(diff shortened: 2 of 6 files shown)


mrooney:
Makes sense, thanks! Is there any reason to have the ISO8601_FMT and DATE_FMT? They don't seem to be used so my suggestion would be to move the EXCEL_FMT definition right above the place where it is used and remove the others.

felciano:
Just for convenience in case anyone wanted a different format. I wasn't sure what the design intention was, so ISO might be preferred over pragmatic concerns like Excel compatibility. I'm fine removing them if that would help. 

I also saw that the nightly build failed due to code convention errors. I'm not sure how to get those builds running against my branch but happy to fix if that would be helpful.

mrooney:
Thanks for the pep8 cleanup! Overall looks great :) I think instead of having random constants in the middle of the file, it might be friendlier if the `presentable_format` argument took a string, either "EXCEL", "ISO8601", or "DATE", and then used a dictionary to populate those. That way you can just look at the function and see the options, and there aren't constants hanging randomly around unused. If that sounds good, then I'll merge and cut a new release. Thanks @felciano for your time and contribution!

felciano:
Sure -- that sounds like a cleaner encapsulation. Looking forward to seeing
the new release!

mrooney:
Thanks, released as 1.25!

PR #107, opened 2017-08-10
str and error concatenation TypeError + small grammar fix

dherg:
Fixes `TypeError: cannot concatenate 'str' and 'error' objects` being printed instead of the RuntimeError message and original error as desired. This bug occurs when webdriver raises a URLError (for example `[Errno 61] Connection refused`).

Also made a tiny grammar fix in the Readme.


mrooney:
Thank you! Merged and released as 1.26

Issue #105, opened 2017-08-04
'NoneType' object has no attribute '__getitem__'


musicmovesme:
Getting this right after login:
Traceback (most recent call last):
  File "mintapi/api.py", line 769, in <module>
    main()
  File "mintapi/api.py", line 708, in main
    mint = Mint.create(email, password, ius_session=options.session, thx_guid=options.thx_guid)
  File "mintapi/api.py", line 74, in create
    mint.login_and_get_token(email, password, ius_session, thx_guid)
  File "mintapi/api.py", line 135, in login_and_get_token
    session_cookies = self.get_session_cookies(**data)
  File "mintapi/api.py", line 199, in get_session_cookies
    'ius_session': driver.get_cookie('ius_session')['value'],
TypeError: 'NoneType' object has no attribute '__getitem__'

dsinger51:
Same issue for me.

dherg:
Looks like `driver.get_cookie()` isn't getting the cookies for ius_session or thx_guid. I think this has to do with the domains for ius_session and thx_guid being accounts.intuit.com and pf.intuit.com respectively, but `get_cookie()` is called while on the domain mint.intuit.com, so it only sees cookies with domains mint.intuit.com and .intuit.com.

A workaround to get ius_session is to accounts.intuit.com and get the cookie there:
`driver.get('http://accounts.intuit.com')`
`ius_session_cookie = driver.get_cookie('ius_session')['value']`

Doesn't seem to be as simple as going to pf.intuit.com to get the thx_guid cookie, however... not sure of a solution there.

mrooney:
Thanks, fixed and released as 1.27!

PR #108, opened 2017-08-14
Change domains before requesting ius_session and thx_guid cookies

dherg:
A fix for #105, navigate to accounts.intuit.com before requesting ius_session cookie and pf.intuit.com before requesting thx_guid cookie. Let me know if this makes sense or if there's a better way to do it.

Also, quick question: is the thx_guid cookie actually being used? I see that ius_session is added to the url in `login_and_get_token()`, but I don't see anywhere the thx_guid cookie is used.


mrooney:
Works for me, thanks!

PR #115, opened 2017-09-20
Fix get_transactions_json when start_date predates all transactions

jprouty:
If start_date is before the earliest/first transaction the if start_date block errors out, since txns is empty. To handle this, simply re-order the blocks.


mrooney:
Thank you, merged and released as 1.28!

PR #120, opened 2017-09-29
Use seleniumrequests to perform all requests.

jprouty:
This PR is in response to Intuit changing their auth cookies recently, and should support a more robust to future change and browser like behavior as it performs all requests in a real browser and should always be up to date based on production Mint web client.


mintapi/api.py
```diff
@@ -130,14 +135,9 @@ def login_and_get_token(self, email, password, ius_session, thx_guid):  # {{{
...
+        if not session_cookies:
             session_cookies = self.get_session_cookies(**data)
-        else:
-            session_cookies = {
-                'ius_session': ius_session,
-                'thx_guid': thx_guid
-            }
         self.cookies.update(session_cookies)
 
         self.get('https://pf.intuit.com/fp/tags?js=0&org_id=v60nf4oj&session_id=' + self.cookies['ius_session'])
```
mitch-lindsay: This isn't valid anymore because that cookie isn't defined.

mintapi/api.py
```diff
@@ -194,26 +194,13 @@ def get_session_cookies(self, username, password):
         while not driver.current_url.startswith('https://mint.intuit.com/overview.event'):
             time.sleep(1)
 
-        # get ius_session cookie by going to accounts.intuit.com
+        # Get session cookies by going to accounts.intuit.com
         driver.get("http://accounts.intuit.com")
```
mitch-lindsay: While testing this, I'm noticing that this request no longer provides the data we would expect.  It simply renders the login form again regardless of the success/failure of the login.

mintapi/api.py
jprouty: Interesting; I was getting this cookie back in my response. Are you not seeing it?

mintapi/api.py
jprouty: Right; I'm not really sure I follow the background behind the change (see https://github.com/mrooney/mintapi/commit/85892cfb7d9b2b5a9d112825fb789d683e1d7161#diff-4117f091b0b8c6d18c2469a6f2452c55) to have this additional nav before grabbing cookies.

https://mint.intuit.com/overview.event should have all the cookies already at this point. Not sure where/how a different domain would address this.

My suggestion would be to remove this extra "redirect", unless I'm missing context. @dherg

mintapi/api.py
mitch-lindsay: Unfortunately not.  That line consistently raised a KeyError for me.
`KeyError: "name='ius_session', domain=None, path=None"`
I wonder if cookies are stored over subsequent sessions of Chrome webdriver?

mintapi/api.py
jprouty: Upon further review, it looks like the ius_session cookie is only after the accounts.intuit.com visit. I was trying to trace the new auth flow with inspector. It seems that the value of ius_session is still sent out to pf.intuit.com/../clear.png, despite never being sent to the client in that form. My guess is there is an alternative representation for the ius_session cookie.

I'm curious why you're not getting the ius_session cookie. Maybe a sleep or some sort is required before grabbing the cookies after the accounts.intuit.com visit?

felciano:
Hi @jprouty I tried checking this out into my environment and running it and am now able to successfully connect and download `--transactions` as well as `--net-worth`. However when I run it with the --accounts switch it outputs `null`. Is the latter working for you?

mintapi/tests.py
```diff
@@ -50,7 +50,9 @@ def tearDown(self):  # noqa
         mintapi.api.Mint = self._Mint
 
     def test_accounts(self):
-        accounts = mintapi.get_accounts('foo', 'bar', ius_session='baz')
+        accounts = mintapi.get_accounts('foo', 'bar', session_cookies={
```
felciano: @jprouty it looks like you are testing a call to `get_accounts` with a different method signature, passing in the new `session_cookies` parameter. Shouldn't this also be updated in the main `api.py` file for handling the `--accounts` switch? I think it is here: https://github.com/jprouty/mintapi/blob/ce19db9b83c38d9d899256eef57c9db2cd7fe542/mintapi/api.py#L728

jprouty: Based on a quick read, I'm pretty sure this flow is unchanged from before. The code you're commenting on is the module function, not the class method on Mint. The main script first constructs a Mint object (with session cookies if given, or gets them from Mint) then does a get_accounts on that client instance.

mintapi/api.py
mitch-lindsay: Can't explain it but I'm not seeing this anymore.  Pretty sure you can ignore this issue.

jprouty: I was also seeing some spurious behavior. I have a suspicion it's due to timing (not waiting for webdriver to finish loading the response and therefore setting the ius_session)... Looking into reproducing this now.

jprouty:
@mitch-lindsay @felciano Please review this approach, as suggested by Mitch and verified by @bmasterc

mrooney:
Thanks! How's this working for anyone else? I like the idea, though my main worry is that the tests file is deleted. We should be able to mock some of these things and test at least some of the most basic options (accounts and transactions).

felciano:
Hi @mrooney I tested @jprouty's selenium approach on https://github.com/jprouty/mint-amazon-tagger and it seems to work reliably. What would the next steps be for getting this approach further evaluated for integration into the mainline `mintapi`?

jprouty:
I'll add back the test and let's give it a more formal review. Otherwise, I think a lot of people have been broken for a while (not sure what our alternative is at this point).

jprouty:
Okay. @felciano @mrooney  I've restored the existing test case but this time using the unittest.mock library. Let me know what you think.

felciano:
Thanks @jprouty. I am trying to get this pull request merged locally, but the github web UI doesn't show me the "command line instructions" as described at https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-active-pull-request-locally. This may be because I don't own the repo. Do you happen to know of alternate instructions / best practices for contributing to GitHub projects by reviewing pull requests w/out being a committer?

Alternately, is this pull request just your master at https://github.com/jprouty/mintapi? If so I can just test that out.

jprouty:
It is indeed my master. Feel free to give that a go.

jprouty:
You can tell at the top of this page where it says jprouty:master.

felciano:
That that's what I figured. Somehow that doesn't make sense to me, since I assume I'll actually be testing your full master, including this patch plus any other changes you've made. I guess I was looking for a workflow that allows contributors to selectively merge the changes scoped to a pull request directly.

Either way, I've now tested your branch for 24 hours and it has worked every time, so from my perspective this is a definite improvement in stability / reproducibility.

jprouty:
@mrooney thoughts? Let me know where you'd like to take this.

@felciano Thanks for testing it out! Re: git. You can tell what's being merged:
`jprouty  wants to merge 25 commits into **mrooney:master** from **jprouty:master**`

When you create a pull request, you selecting which repo and branch to merge. Github then commutes if they have conflicts (this one doesn't, thus the green check) and how many commits there are between those two branches. In this case, my master has 25 commits in this change. In the case of a real world development workflow, I'd probably development this change on a different branch than master, and keep master synced with upstream (which would be mrooney's version). But since I only have 1 change going, and my whole fork for the purpose of said change, I just did it on master itself. Hope that helps!

jprouty:
Thanks @felciano. Fixed!

mrooney:
Checking it out now, thanks! I'm noticing that the Chrome window always seems to open a version/updating tab, and the whole thing never seems to close once it successfully gets the balances. Is that what you are seeing as well?

jprouty:
@mrooney My guess is that your version of chromium (or chrome depending on how you setup selenium) is not current, which is spawning a new window trying to update? What if you run the full update in that window? Or maybe update the chromium-webdriver package.

I've never see this problem. Can you verify the mint browser window is closing? That should be handled via the `atexit` and `Mint#close`.

mrooney:
Thanks! I like the general approach. A couple thoughts:

1. Doing a `brew upgrade chromedriver` worked, but it seems a bummer that this will probably happen regularly. I wonder if there is a way to disable the upgrade check and tab?
2. The mint browser window does close, but if you use it from a longer-running app that imports mintapi, it doesn't close until the app closes (which could be hours or days), so that's a bit unexpected.
3. I'm noticing that I get asked to get a code emailed or texted each time. Is that also what you are experiencing? Is there a way to specify a cookie/cache location for mintapi's usage so it remembers us?

jprouty:
1) Great! There is most likely a way to disable it (by passing in some cmd args when launching the webdriver). However, I'd highly recommend we don't; this scrapper is for financial data and if your webdriver goes for a long time without update, that's just not good. Maybe we address that annoyance with a note in the README?

2) Yup, not sure if there's a good way around this... maybe we could minimize the window after successful login? Not sure if we have that level of control.

3) I have not run into that. Only the normal amount of challenges. Are you logging in many times in a short period?

mrooney:
Got it. Well, it is definitely an improvement. I'll merge it!

jprouty:
Woot! Thanks for the merge. Hopefully people find it a welcome fix 👍

jprouty:
@mrooney Any chance we can bump the version on pypi? 👍

mrooney:
Thanks for the reminder @jprouty ! Released as 1.29 to pypi :)

PR #124, opened 2017-12-28
Run headless in docker container

kkochubey1:
Just an example on how to run it from docker container in headless mode. I do not think is should be part of your original repository. But it can be there as an example.

Is there plan to support new chrome --headless mode?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,57 @@
+FROM ubuntu:xenial
+
+RUN apt-get update -yq
+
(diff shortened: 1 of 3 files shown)


jorythompson:
Running headless is pretty easy.  Just need to pass the following to the driver:
`driver_options.add_argument("headless")`

mrooney:
Thanks for this example, I'll close this as we now have --headless :)

PR #138, opened 2018-05-01
Update setup.py

ak2k:
Worth including pandas, since appears to be a requirement for some of the json/csv file output?


mrooney:
Seems reasonable, thanks!

PR #141, opened 2018-05-22
inherit Mint from object

flavioribeiro:
just so I can do https://github.com/flavioribeiro/update-my-mint/blob/master/mint.py#L15


mrooney:
Thanks!

PR #153, opened 2018-09-09
Updated install instructions for chromedriver

felciano:
chromedriver has migrated from homebrew/core to homebrew/cask


mrooney:
Thanks!

PR #156, opened 2018-10-06
Don't hang if some accounts need attention

dmulder:
The "refresh completed" message is slightly different if some accounts need attention. Instead it says "Account refresh complete." then mentions how many accounts need attention. If we trim the message, and instead check if innerHTML *contains* the message (instead of equals), then both situations are handled.


mrooney:
Thanks, great catch!

Issue #147, opened 2018-07-08
mint.get_accounts() does not work after successful login


smnlng:
Running python 3.6 with selenium installed. After the initial code 

`import mintapi
mint = mintapi.Mint("email", "password")`

There is a successful login to Mint using a new instance of Chrome. However, attempting to use 
`mint.get_accounts()` or `mint.get_transactions_csv(include_investment=False)`

Creates a new window to http://127.0.0.1:51907/ and nothing happens. Chromedriver reports the following:

` [3020:6080:0707/232340.899:ERROR:install_util.cc(589)] Unable to create registry key HKLM\SOFTWARE\Policies\Google\Chrome for reading result=2`

`DevTools listening on ws://127.0.0.1:52065/devtools/browser/<snip>
[3020:6080:0707/232342.468:ERROR:gpu_process_transport_factory.cc(1017)] Lost UI shared context. `

Any suggestions?

jaron-l:
This is a duplicate of #245, which has a fix that has been merged. Let us know if you continue to have issues.

PR #159, opened 2018-10-20
Fix for defect #147 mint.get_accounts() does not work after successfu…

salmoni88:
This is my first PR.  Included a fix to parse_float and included unit tests.


mrooney:
Thanks, this is fantastic!

PR #160, opened 2018-10-21
Add session-path argument

apiology:
Allow Chrome session to be saved between invocations, which will prevent 2FA from being prompted for multiple times in headless cloud environments where the user can't log in by hand in a browser.

This should help with some of the challenges raised in #137 , as it will let someone either log in once and create a working session, or just ship a valid session over to the cloud environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,12 @@ Ensure you have Python 2 or 3 and pip (`easy_install pip`) and then:
 
 ```shell
 pip install mintapi
-```
-
-If you do not want to manually find and provide your Mint session cookies, as described below, then please also install `selenium` and `chromedriver` (version 59+ if you want to use headless mode):
-```shell
-pip install selenium
 brew cask install chromedriver # or sudo apt-get install chromium-chromedriver on Ubuntu/Debian
 ```
+
+Note that chromedriver must be version 59+ if you want to use headless mode.
+
+```shell
 
 ## Usage
 
@@ -33,9 +32,13 @@ make calls to retrieve account/budget information.  We recommend using the
                        # if mintapi detects an MFA request, it will trigger the requested method and prompt on the command line.
     headless=False,  # Whether the chromedriver should work without opening a
                      # visible window (useful for server-side deployments)
-    mfa_input_callback=None  # A callback accepting a single argument (the prompt)
-                             # which returns the user-inputted 2FA code. By default
-                             # the default Python `input` function is used.
+    mfa_input_callback=None,  # A callback accepting a single argument (the prompt)
+                              # which returns the user-inputted 2FA code. By default
+                              # the default Python `input` function is used.
+    session_path=None, # Directory that the Chrome persistent session will be written/read from.
+                       # To avoid the 2FA code being asked for multiple times, you can either set
+                       # this parameter or log in by hand in Chrome under the same user this runs
+                       # as.
   )
 
   # Get basic account information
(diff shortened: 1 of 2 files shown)


mintapi/api.py
```diff
@@ -95,12 +96,20 @@ def get_web_driver(email, password, headless=False, mfa_method=None,
         chrome_options.add_argument('disable-dev-shm-usage')
         chrome_options.add_argument('disable-gpu')
         # chrome_options.add_argument("--window-size=1920x1080")
+    if session_path is not None:
```
mrooney: I think it would be really great if this could be default behavior, as I imagine most folks would want it and it might be hard to discover. I would suggest picking some sensible default (home / user dir perhaps) but allowing it to be disabled. Ie, "--session-path=None" will disable this feature, but you can specify a directory and if you don't you get a sensible default.

mrooney: I think it would be really great if this could be default behavior, as I imagine most folks would want it and it might be hard to discover. I would suggest picking some sensible default (home / user dir perhaps) but allowing it to be disabled. Ie, "--session-path=None" will disable this feature, but you can specify a directory and if you don't you get a sensible default.

mrooney: I think it would be really great if this could be default behavior, as I imagine most folks would want it and it might be hard to discover. I would suggest picking some sensible default (home / user dir perhaps) but allowing it to be disabled. Ie, "--session-path=None" will disable this feature, but you can specify a directory and if you don't you get a sensible default.

mrooney: I think it would be really great if this could be default behavior, as I imagine most folks would want it and it might be hard to discover. I would suggest picking some sensible default (home / user dir perhaps) but allowing it to be disabled. Ie, "--session-path=None" will disable this feature, but you can specify a directory and if you don't you get a sensible default.

mrooney: I think it would be really great if this could be default behavior, as I imagine most folks would want it and it might be hard to discover. I would suggest picking some sensible default (home / user dir perhaps) but allowing it to be disabled. Ie, "--session-path=None" will disable this feature, but you can specify a directory and if you don't you get a sensible default.

README.md
```diff
@@ -7,13 +7,12 @@ Ensure you have Python 2 or 3 and pip (`easy_install pip`) and then:
...
 ```shell
 pip install mintapi
+brew cask install chromedriver # or sudo apt-get install chromium-chromedriver on Ubuntu/Debian
 ```
 
-If you do not want to manually find and provide your Mint session cookies, as described below, then please also install `selenium` and `chromedriver` (version 59+ if you want to use headless mode):
+Note that chromedriver must be version 59+ if you want to use headless mode.
+
 ```shell
-pip install selenium
```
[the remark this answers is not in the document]
apiology: Correct, it's handled by the default install.  The `install_requires` line [here](https://github.com/mrooney/mintapi/blob/master/setup.py#L23) sets up transitive dependencies; pip installs those as well when being asked to install mintapi.

mintapi/api.py
apiology: Sounds good - will look into it.

mrooney:
Thanks @apiology ! I'm not seeing any new commits, is it possible it needs a click to update the PR here?

vinceatbluelabs:
Sorry for the confusion @mrooney - the new code is here, the commits were squashed.

apiology:
Sorry for the confusion @mrooney - the new code is here, the commits were squashed.

apiology:
Hi @mrooney - did you get a chance to take a look at the code I added?

PR #166, opened 2019-02-01
Automate mfa-code retrieval from email via IMAP

tsdg112:
This change allows mintapi to run autonomously, even when Mint requires MFA.  If mfa-method is email and your email host provides IMAP access, you can specify your IMAP login details.  This will automate the retrieval of the MFA code from your email and entering it into Mint.

There is a also a small change to allow an account to be specified when calling mint.get_transactions_json and _csv.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,15 +69,6 @@ make calls to retrieve account/budget information.  We recommend using the
   mint.get_transactions_csv(include_investment=False) # as raw csv data
   mint.get_transactions_json(include_investment=False, skip_duplicates=False)
 
-  # Get transaction data separately for each account
-  # adding the key "id" identifying the acount to each transaction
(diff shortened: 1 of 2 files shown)


mrooney:
Thanks, this looks very thorough and impressive! Do you think this is still needed with the recent addition of storing sessions to avoid repeated MFAs? And if so, would you mind resolving the current conflicts? I'm happy to review and merge if so! Thanks for your time and contribution.

tsdg112:
My experience is that with session storage, Mint will work for a while (days or weeks) without requiring MFA.  But then, out of the blue, it requires MFA.  I see the same thing happen when accessing Mint from a browser, so this is probably not an issue with the API.

mrooney:
Cool, merged and released as 1.38, thanks for your contribution!

PR #163, opened 2018-12-30
Adding ability to get credit score.

willbiddy:
Gets the credit score from Mint. Includes a command line flag and updated documentation.
Fixes issue #142

PR #169, opened 2019-03-02
Full Credit Report API

hshore29:
I was playing around with Mint, and figured out how their Credit Report API works - a more robust solution to getting the credit score, as it also pulls in all prior credit reports (not just the current one), and metadata / history about all of the metrics that contribute to the score.

The Credit Report API is on a different domain (https://credit.finance.intuit.com), and uses a different authorization scheme. There are 4 data endpoints, and an additional metadata endpoint.

**Authorization**
The Credit API uses an authorization header and API key. The API key is saved to the window scope in JavaScript on page load, and can be retrieved by returning `window.MintConfig.browserAuthAPIKey`. Then, pass the token in an authorization header with the value `Intuit_APIKey intuit_apikey={api_key}` when requesting data from any of the end points. All of these support `GET` requests.

**Endpoints**
- **/v1/creditreports**: this takes one parameter (limit=#), and returns up to that many credit reports. Mint loads 2 by default, but they store way more - at least 8, but potentially every report since they switched to TransUnion in ~2017.
- **/v1/creditreports/0/inquiries**: full list of credit inquiries from the latest credit report.
- **/v1/creditreports/0/tradelines**: full list of credit accounts, and their details from the latest credit report.
- **/v1/creditreports/creditutilizationhistory**: a limited credit utilization history by account, going back about 3 months.

Info about how the credit bands are defined can be found by hitting **/v1/creditscoreproviders/3**, but this also includes a large amount of HTML snippets and styling information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ Run it as a sub-process from your favorite language; `pip install mintapi` creat
 ```shell
     usage: mintapi [-h] [--session-path [SESSION_PATH]] [--accounts]
                    [--budgets] [--net-worth] [--extended-accounts] [--transactions]
-                   [--extended-transactions] [--credit-score] [--start-date [START_DATE]]
+                   [--extended-transactions] [--credit-score] [--credit-report] [--start-date [START_DATE]]
                    [--include-investment] [--skip-duplicates] [--show-pending]
                    [--filename FILENAME] [--keyring] [--headless]
                    [--mfa-method {sms,email}]
@@ -105,6 +105,7 @@ Run it as a sub-process from your favorite language; `pip install mintapi` creat
                             profile.
       --budgets             Retrieve budget information
       --credit-score        Retrieve credit score
+      --credit-report       Retrieve full credit report & history
       --net-worth           Retrieve net worth information
       --extended-accounts   Retrieve extended account information (slower, implies
                             --accounts)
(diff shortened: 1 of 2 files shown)


mrooney:
Thanks, that is super cool! I'm unable to test this myself since I haven't given Mint my SSN, but it looks like a nicely coded and isolated feature so I don't have any problem merging this. Thanks for also updating the README :) I will release this to pip now!

mrooney:
Okay, released as 1.37. Any recommendations on how to better integrate this with https://github.com/mrooney/mintapi/pull/163? Does it make sense to have one switch for a simple integer score and another for a full report or should the score be folded into this ideally?

hshore29:
Awesome!
It would be pretty easy to just return the latest integer score from the credit API, so the #163 functionality could be folded in as an option instead of a separate switch. So instead of --credit-report and --credit-score, you'd have --credit-report and --credit-report --score-only or something like that.
I can code that up and submit another PR if you think that makes sense!

mrooney:
@hshore29 that makes sense. I don't think there's anything necessarily bad about having a credit-score switch, at least so it doesn't break on upgrade, but consolidating the code paths to use yours would be super appreciated if you were willing to contribute that as a PR :D

PR #171, opened 2019-04-02
Bypass add your current mobile number interstitial page

tsdg112:
I occasionally get an interstitial page that says "Let's add your current mobile number" and mintapi fails.   This change clicks the "skip for now" link when that page appears.


mrooney:
Awesome, thanks!

PR #173, opened 2019-04-06
Fix PEP8 violations (lots!), pipe through wait_for_sync, and various login flow fixes/timing issues

jprouty:
Also cleans up a hairy section in json transactions where url params are hand crafted!

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -24,7 +24,7 @@ except ImportError:
 from selenium.common.exceptions import ElementNotVisibleException, NoSuchElementException, StaleElementReferenceException, TimeoutException
 from selenium.webdriver import ChromeOptions
 from selenium.webdriver.common.by import By
-from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support import expected_conditions
 from selenium.webdriver.support.ui import WebDriverWait
 
 from seleniumrequests import Chrome
@@ -34,12 +34,6 @@ try:
     import pandas as pd
 except ImportError:
     pd = None
-
-# Temporary pep8 violation to test Travis
-try:
-    pass
-except:
-    pass
 
 
 def assert_pd():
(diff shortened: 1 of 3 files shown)


mintapi/api.py
```diff
@@ -530,29 +550,32 @@ def get_transactions_json(self, include_investment=False,
...
         offset = 0
         # Mint only returns some of the transactions at once.  To get all of
         # them, we have to keep asking for more until we reach the end.
         while 1:
+            url = MINT_ROOT_URL + '/getJsonData.xevent'
+            params = {
+                'queryNew': '',
+                'offset': offset,
+                'comparableType': '8',
+                'rnd': Mint.get_rnd(),
```
mrooney: This looks great! But what happens to query_options?

mrooney: Ah I see, it ends up as params :)

jprouty: Yup! the requests lib does all the url encoding instead :) Way cleaner and less error prone. This lib was already using params elsewhere, so this is nice to start unifying.

jprouty:
Honestly, I'm not sure what's going on with the pep8 regressions! It was definitely working (a year ago?!) when I made my last PR. I'm happy to look a bit closer. Do you know if the modified travis.yaml modifies the current run? Or only the checked in travis is used?

jprouty:
Oh; also worth mentioning: I verified everything is still working via accounts, budgets, and extended-transactions (since that was modified)

jprouty:
Alright; this latest commit revives it. Guess who broke it 1 year ago... :) it's 'before_script', not 'before-script' :) My bad!

jprouty:
In addition, **I've changed the py3 version to be 3.6**, which runs at the normal speed. For some reason, pandas (and therefore numpy) is freaking out on 3.4 and takes 10x the time the py2.7 and py3.6 test takes. I'd do 3.7, but for some reason that's not available on travis (probably older version of ubuntu or something).

This run illustrates all the versions together: https://travis-ci.org/mrooney/mintapi/builds/516678277

Should be all set for final review. Thanks!

mrooney:
Beautiful, fantastic work, thanks so much for this great contribution :D

jprouty:
My pleasure and apologies again for breaking the Travis CI :)

PR #175, opened 2019-04-25
Move get_credit_score onto Credit API code

hshore29:
Instead of having two ways to get credit scores (scraping the page for the score, or using Mint's credit API for the full report), this moves both onto the credit API. get_credit_score() now wraps get_credit_report(), returning just the score instead of the full credit report object.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -823,14 +823,14 @@ class Mint(object):
             data={'token': self.token},
             headers=JSON_HEADER)
 
-    def get_credit_score(self, limit):
-        credit_report = self.get_credit_report(limit, details=False)
-        # Extract the scores as a list
-        scores = list()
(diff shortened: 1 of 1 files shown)


mrooney:
Thanks, this looks cool! The only issue I see is that it changes our contract from returning an int to returning a list of ints, for the credit score. This means that anyone automating/parsing CLI output will be broken.

Was the one in the UI it was previously scraping the first one or an average? We should consider just returning that, and people have the full method if they want them all or more data. This also better matches the name "credit score" and not "scores". What do you think?

hshore29:
Makes sense to me! The old version would have just gotten the most recent score - so I updated the new get_score function to just request one report, and pull the score out and just return that.

mrooney:
Looks great, thanks!

PR #180, opened 2019-06-06
Add wait_for_sync_timeout

tsdg112:
Wait_for_sync waited 300 seconds for Mint for sync with the institution.  I added wait_for_sync_timeout to specify the number of seconds to wait.  The default is 300 seconds.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -572,7 +572,7 @@ class Mint(object):
             # transactions as well.  Otherwise they are skipped by
             # default.
             if id > 0 or include_investment:
-                params['id'] = id
+                params['accountId'] = id
             if include_investment:
                 params['task'] = 'transactions'
             else:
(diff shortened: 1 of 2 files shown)


tsdg112:
A typo in commit e069366a299c7ff22167184728692a8b4a8c644a broke retrieving transactions for a specified account.  The second commit here fixes it.

mintapi/api.py
```diff
@@ -568,7 +572,7 @@ def get_transactions_json(self, include_investment=False,
             # transactions as well.  Otherwise they are skipped by
             # default.
             if id > 0 or include_investment:
-                params['id'] = id
+                params['accountId'] = id
```
mrooney: Looks good overall. What is this change for?

mrooney:
Thanks, looks nice, just added one question :)

mintapi/api.py
tsdg112: This was the typo in commit e069366 that broke retrieving transactions for a specified account. This fixes it.

mrooney:
Thank you for your contribution!

PR #186, opened 2019-07-12
fix mfa issue with gmail

boriza:
email mfa was not working for me due to this code. This changed fixed the issue


tsdg112:
This commit removes the code that ensures that the email MFA message is not stale (less than 3 minutes old).  This is not a good idea because mintapi might try to use an older email with a previous MFA code.

Why is gmail returning bad timestamps?  Perhaps you did not set the timezone correctly in gmail.

mrooney:
Thanks for the contribution, however I'm going to close this due to lack of response from the opener. Feel free to create a new PR addressing the comments if this still applies. Thanks!

PR #192, opened 2019-11-23
Add get_invests_json.  Retrieves brokerage holdings (and transactions)

tsdg112:
Adds a new function, get_invests_json.  It retrieves investment holdings from brokerages.  Transactions, broken down by holding, are also retrieves.


mrooney:
Thanks for your contribution!

PR #199, opened 2020-01-10
Add soft token support

chino:
Just a quick hack that seems to work.

This adds support for using "authenticator app" option instead of sms/email.  When the user enables this they are given a token that they save and provide as an argument.

I could add keyring support for it as well if desired.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -255,7 +255,7 @@ def get_stable_chrome_driver(download_directory=os.getcwd()):
     return local_executable_path
 
 
-def get_web_driver(email, password, headless=False, mfa_method=None,
+def get_web_driver(email, password, headless=False, mfa_method=None, mfa_token=None,
                    mfa_input_callback=None, wait_for_sync=True,
                    wait_for_sync_timeout=5 * 60,
                    session_path=None, imap_account=None, imap_password=None,
@@ -324,8 +324,9 @@ def get_web_driver(email, password, headless=False, mfa_method=None,
         driver.implicitly_wait(1)  # seconds
         try:
             if mfa_method == 'soft-token':
+                import oathtool
                 mfa_token_input = driver.find_element_by_id('ius-mfa-soft-token')
-                mfa_code = (mfa_input_callback or input)("Please enter your 6-digit MFA code: ")
+                mfa_code = oathtool.generate_otp(mfa_token)
                 mfa_token_input.send_keys(mfa_code)
                 mfa_token_submit = driver.find_element_by_id('ius-mfa-soft-token-submit-btn')
                 mfa_token_submit.click()
@@ -427,7 +428,7 @@ class Mint(object):
     driver = None
     status_message = None
 
-    def __init__(self, email=None, password=None, mfa_method=None,
+    def __init__(self, email=None, password=None, mfa_method=None, mfa_token=None,
                  mfa_input_callback=None, headless=False, session_path=None,
                  imap_account=None, imap_password=None, imap_server=None,
                  imap_folder="INBOX", wait_for_sync=True, wait_for_sync_timeout=5 * 60,
@@ -436,6 +437,7 @@ class Mint(object):
         if email and password:
             self.login_and_get_token(email, password,
                                      mfa_method=mfa_method,
+                                     mfa_token=mfa_token,
                                      mfa_input_callback=mfa_input_callback,
                                      headless=headless,
                                      session_path=session_path,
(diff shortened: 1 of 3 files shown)


mintapi/api.py
```diff
@@ -15,6 +15,7 @@
 import email
 import email.header
 import sys  # DEBUG
+import oathtool
```
mrooney: Can we do this import inside of the `if mfa_method == 'soft-token':` so that someone pulling the new code doesn't get broken without installing it?

mintapi/api.py
```diff
@@ -213,7 +214,7 @@ def get_web_driver(email, password, headless=False, mfa_method=None,
     if session_path is not None:
         chrome_options.add_argument("user-data-dir=%s" % session_path)
 
-    driver = Chrome(chrome_options=chrome_options, executable_path="%s" % executable_path)
+    driver = Chrome(options=chrome_options, executable_path="%s" % executable_path)
```
mrooney: Is this just a change in recent Chrome where chrome_options is older / deprecated?

mintapi/api.py
```diff
@@ -1130,8 +1133,12 @@ def main():
     cmdline.add_argument(
         '--mfa-method',
         default='sms',
-        choices=['sms', 'email'],
+        choices=['sms', 'email', 'soft-token'],
         help='The MFA method to automate.')
+    cmdline.add_argument(
+        '--mfa-token',
+        default=None,
+        help='The MFA soft-token to pass to oathtool.')
```
mrooney: If mfa-token logically implies the method of soft token, I'm fine with making that implicit. Your call!

chino: Right now you would have to pass:
```
--mfa-method soft-token --mfa-token $token
```

We could change it to just:
```
--mfa-token $token
```

Perhaps another idea is something like this if it's possible:
```
--mfa-method soft-token=$token
```

I'm not sure what is the most idiomatic style for python or this library.

mrooney: I think that's fine if that seems simplest! If the mfa-method was implicit if you passed a token that would also be fine.

setup.py
```diff
@@ -16,7 +16,7 @@
     author='Michael Rooney',
     author_email='mrooney.mintapi@rowk.com',
     url='https://github.com/mrooney/mintapi',
-    install_requires=['future', 'mock', 'requests', 'selenium-requests', 'xmltodict', 'pandas>=1.0', 'selenium'],
+    install_requires=['future', 'mock', 'requests', 'selenium-requests', 'xmltodict', 'pandas>=1.0', 'selenium', 'oathtool'],
```
mrooney, line 19: Since you only import it if that argument is passed, technically it doesn't require it, but I guess why not?

mintapi/api.py
```diff
@@ -324,8 +324,9 @@ def get_web_driver(email, password, headless=False, mfa_method=None,
         driver.implicitly_wait(1)  # seconds
         try:
             if mfa_method == 'soft-token':
+                import oathtool
                 mfa_token_input = driver.find_element_by_id('ius-mfa-soft-token')
-                mfa_code = (mfa_input_callback or input)("Please enter your 6-digit MFA code: ")
+                mfa_code = oathtool.generate_otp(mfa_token)
```
mrooney, line 329: How is it generating a OTP without input? This seems too magical, I must be missing something. Do you configure oathtool at the system / user level outside of mintapi or something?

setup.py
chino: You're right, I just didn't think it through after making the other change.

mintapi/api.py
chino: The token passed in generates another time based token that is passed to the api. It's similar to a passwordless ssh key in that you're not promoted for a password to access the key. 

The main user password is still required though that's separate.

I could take a look how to password protect the key and add that. 

Most things like Google Authenticator work similarly where you add a token to it either manually or by a qr code and then it's up to the app to optionally protect the access via passcode or bio fingerprint and so forth.

chino: Note also that a user calling the script has full freedom to protect the token using things like `git-crypt`, `security` tool on osx for keychain, or anything they desire. It doesn't have to be the job of this tool to implement that.

kandluis: I tested this change and works as specified.

@mrooney: Getting it to work involves setting up your Minta account to use 2FA with Google Authenticator. In addition to the QR code, it also provides a token which when passed in to this took generates the OTP required to login.

setup.py
chino: Hey, @mrooney I think you wanted to remove this before merge?

mrooney: > Hey, @mrooney I think you wanted to remove this before merge?

It's okay, I think it is fine there!

mrooney:
Thanks, merged and released to pypi as 1.44!

Issue #201, opened 2020-02-19
Add ability to use chromedriver in PATH


jaron-l:
The current [get_web_driver](https://github.com/mrooney/mintapi/blob/4bf3f6fe6b90005b4e2b0b63c690786b869bfc4d/mintapi/api.py#L188-L216)  function doesn't check the PATH to see if chromedriver is there. It only checks the current working directory for chromedriver and then attempts to download chromedriver version 2.41 to be used.

It would be easy for me to submit a pull request to check for chromedriver in path and the current directory but I think this is related to a similar issue:

I think it should be considered whether the app should be attempting to get chromedriver for the user at all. I'm not sure if this is causing other issues but if the app will only use version 2.41 of chromedriver published in 2018, then it may cause issues in the future. To me, several of the comments in #95 seem to suggest issues with the chromedriver version not working and none of them seem to have ever been resolved except for the one post where he says he manually put in the path to his chromedriver. The easy way out would be to require the user to provide the chromedriver in PATH or in the current directory.

For documentation purposes, I briefly researched how hard a more robust method might be to automatically find the correct chromedriver version. It would probably require a operating system check and then either run [this command in linux](https://serverfault.com/a/646402) or [a command referenced here in windows](https://bugs.chromium.org/p/chromium/issues/detail?id=158372#c15) that would determine what chrome/chromium version is installed and then download the proper chromedriver according to [these instructions](https://sites.google.com/a/chromium.org/chromedriver/downloads/version-selection). I know that's a lot but if it seems like a good direction to go in, I could submit a pull request.

jprouty:
Wow, crazy coincidence, I was going to open an issue right this second around a similar issue. I would suggest moving to the chromedriver-py package, which will allow users to selectively upgrade/downgrade the exact version as they please, and will make distribution/caching of versions easier.

I'm going to send a pull request over to get the ball going.

jprouty:
For context, the main _driver_ for me, is that I have users of my project (jprouty/mint-amazon-tagger) that are occasionally running into issues ([issue44](https://github.com/jprouty/mint-amazon-tagger/issues/44)) that appear to be version compat issues.

mrooney:
Thanks, I'll close in favor of #202

PR #202, opened 2020-02-19
Update chromedriver fetch logic to use the latest version from the stable channel (instead of a hard coded version). Related, allow by cli/api the ability to use chromedriver on PATH

jprouty:
See #201 for context. Turns out that using the latest and greatest (major release 81) doesn't work with older versions. So instead of using something from beta channel, most users will stick to something on stable. If a user has an out of date chrome browser, they may run into issues, and should update their browser before using any fintech websites anyway.

I looked into the various chromedriver pypi packages, and it seems that all the ones that either bundle the binary in the package, or download chomedriver and install on the path at module install time are all grabbing the beta latest, not the stable latest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Ensure you have Python 2 or 3 and pip (`easy_install pip`) and then:
 pip install mintapi
 ```
 
-`mintapi` scrapes Mint.com by navigating a Chrome browser (or Chromium) just as a human would. Once logged in, the API allows programatic access to various Mint REST APIs. Selenium/WebDriver is used to accomplish this, and specifically, ChromeDriver under the hood. `mintapi` will download the latest stable release of chromedriver, unless --use_chromedriver_on_path is given. **NOTE: You must have [Chrome](https://www.google.com/chrome/) or [Chromium](https://www.chromium.org/getting-involved/dev-channel/) installed, on the `stable` track, and be up-to-date!**
+`mintapi` scrapes Mint.com by navigating a Chrome browser (or Chromium) just as a human would. Once logged in, the API allows programatic access to various Mint REST APIs. Selenium/WebDriver is used to accomplish this, and specifically, ChromeDriver under the hood. `mintapi` will download the latest stable release of chromedriver, unless --use_chromedriver_on_path is given. **NOTE: You must have [Chrome](https://www.google.com/chrome/) or [Chromium](https://www.chromium.org/getting-involved/dev-channel/) installed, on the `stable` track, and be up-to-date!** If you run into a `SessionNotCreatedException` about "ChromeDriver only supports Chrome version XX", you need to [update Chrome](https://support.google.com/chrome/answer/95414).
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Ensure you have Python 2 or 3 and pip (`easy_install pip`) and then:
 pip install mintapi
 ```
 
-`mintapi` scrapes Mint.com by navigating a Chrome browser (or Chromium) just as a human would. Once logged in, the API allows programatic access to various Mint REST APIs. Selenium/WebDriver is used to accomplish this, and specifically, ChromeDriver under the hood. `mintapi` will download the latest stable release of chromedriver, unless --use_chromedriver_on_path is given. **NOTE: You must have [Chrome](https://www.google.com/chrome/) or [Chromium](https://www.chromium.org/getting-involved/dev-channel/) installed, and on the `stable` track.**
+`mintapi` scrapes Mint.com by navigating a Chrome browser (or Chromium) just as a human would. Once logged in, the API allows programatic access to various Mint REST APIs. Selenium/WebDriver is used to accomplish this, and specifically, ChromeDriver under the hood. `mintapi` will download the latest stable release of chromedriver, unless --use_chromedriver_on_path is given. **NOTE: You must have [Chrome](https://www.google.com/chrome/) or [Chromium](https://www.chromium.org/getting-involved/dev-channel/) installed, on the `stable` track, and be up-to-date!**
 
 ## Usage
 
@@ -85,7 +85,7 @@ make calls to retrieve account/budget information.  We recommend using the
 
   # Get net worth
   mint.get_net_worth()
-  
+
   # Get credit score
   mint.get_credit_score()
 
@@ -94,7 +94,7 @@ make calls to retrieve account/budget information.  We recommend using the
 
   # Get investments (holdings and transactions)
   mint.get_invests_json()
-  
+
   # Close session and exit cleanly from selenium/chromedriver
   mint.close()
 
--- a/README.md
+++ b/README.md
@@ -7,11 +7,9 @@ Ensure you have Python 2 or 3 and pip (`easy_install pip`) and then:
 
 ```shell
 pip install mintapi
-brew cask install chromedriver # or sudo apt-get install chromium-chromedriver on Ubuntu/Debian
 ```
 
-Note that chromedriver must be version 59+ if you want to use headless mode. If not installing via pip,
-make sure to install the `install_requires` dependencies from setup.py yourself.
+`mintapi` does it scraping by driving an instance of Chrome (or chromium) and navigating, as a human user would, thought the login flows. Once logged in, the API allows programatic access to various Mint REST APIs. Selenium/WebDriver is used to accomplish this, and specifically, ChromeDriver under the hood. `mintapi` will download the latest stable release of chromedriver, unless --use_chromedriver_on_path is given. **NOTE: You must have [Chrome](https://www.google.com/chrome/) or [Chromium](https://www.chromium.org/getting-involved/dev-channel/) installed, and on the `stable` track.**
 
 ## Usage
 
@@ -60,6 +58,8 @@ make calls to retrieve account/budget information.  We recommend using the
     imap_folder='INBOX',  # IMAP folder that receives MFA email
     wait_for_sync=False,  # do not wait for accounts to sync
     wait_for_sync_timeout=300,  # number of seconds to wait for sync
+	use_chromedriver_on_path=False,  # True will use a system provided chromedriver binary that
+	                                 # is on the PATH (instead of downloading the latest version)
   )
 
   # Get basic account information
@@ -149,6 +149,9 @@ Run it as a sub-process from your favorite language; `pip install mintapi` creat
       --keyring             Use OS keyring for storing password information
       --headless            Whether to execute chromedriver with no visible
                             window.
+	  --use-chromedriver-on-path
+	  						Whether to use the chromedriver on PATH, instead of
+              			  	downloading a local copy.
       --mfa-method {sms,email}
                             The MFA method to automate.
       --imap-account IMAP_ACCOUNT
@@ -160,6 +163,7 @@ Run it as a sub-process from your favorite language; `pip install mintapi` creat
       --no_wait_for_sync    Do not wait for accounts to sync
       --wait_for_sync_timeout
                             Number of seconds to wait for sync (default is 300)
+
 
     >>> mintapi --keyring email@example.com
     [
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Ensure you have Python 2 or 3 and pip (`easy_install pip`) and then:
 pip install mintapi
 ```
 
-`mintapi` does it scraping by driving an instance of Chrome (or chromium) and navigating, as a human user would, thought the login flows. Once logged in, the API allows programatic access to various Mint REST APIs. Selenium/WebDriver is used to accomplish this, and specifically, ChromeDriver under the hood. `mintapi` will download the latest stable release of chromedriver, unless --use_chromedriver_on_path is given. **NOTE: You must have [Chrome](https://www.google.com/chrome/) or [Chromium](https://www.chromium.org/getting-involved/dev-channel/) installed, and on the `stable` track.**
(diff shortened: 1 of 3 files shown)


jprouty:
One last note I need to mention:
- I tried using the PATH based approach for using chromedriver on the path. I was testing on Ubuntu 19.10 w/ the chromium snap,, which provides both the chromium-browser and chromedriver binaries for Ubuntu. The snap seems to lag what's on the chrome stable channel, which isn't great, and means the minor versions between the chromedriver and google-chrome may likely go out of sync (oh ya, Selenium prefers google-chrome over chromium-browser if both are installed...). Even after I installed the candidate snap so the chromium version lines up with the google chrome version, that version was complaining about the user profile being open in more than one spot (but it definitely was not).

Because of that ^^^ I decided to add a flag around where to use the path at all, and default to _not_ using it. If someone technically inclined knows they want to use the path for chromedriver, they're also probably savvy enough to find and use the new flag.

jaron-l:
I feel like your choice to prefer not to use the chromedriver in PATH only delayed the underlying problem of incompatible versions. As people continue to use this program, their chrome/chromium versions will probably get updated while their chromedriver version will be locked as how mintapi downloaded it. If we want to provide chromedriver for the user, shouldn't there be a chrome version check?

jprouty:
@jaron-l Completely agree. I'll add a check to verify chromedriver --version matches the LATEST_RELEASE from chromedriver. If it doesn't I'll remove chromedriver, and download the latest. If the user isn't on the latest version, they will be forced to update Chrome to match the latest version, which is for the best, especially when dealing with financial websites.

Although it would be neat (and considerably more work to get the path selection smooth on all platforms), I'm afraid trying to find the chrome install and getting version, just to fetch a potentially out of date chromedriver, is extra logic and extra fragility.

jprouty:
@mrooney Please take a look when you have a chance. I've tested on Ubuntu and Win10. This should be a big improvement!

jaron-l:
I think this is a really good compromise. I agree, it probably would be over complicated and prone to error what I originally suggested. It might be a good idea to add some sort of mention of this to the README.md so that people know that it will prefer the latest chromedriver. Just a suggestion.

Also, at line 416 on your version, you should pass in the Mint init value of use_chromedriver_on_path to login_and_get_token so that the option is persisted down. I tried to edit it directly on your version but I don't think it will let me. It only let's me edit code directly around your changes.

jprouty:
@jaron-l Done and done. I meant to update the readme but ran out of gas yesterday ;) Also fixed the init value. @mrooney, mind taking a look?

jprouty:
Also have tested on MacOS w/ latest chrome; all good.

mintapi/api.py
```diff
@@ -59,27 +64,27 @@ def reverse_credit_amount(row):
     return amount if row['isDebit'] else -amount
 
 
-def get_email_code(imap_account, imap_password, imap_server, imap_folder, debug=0, delete=1):
```
mrooney: Won't changing this signature break usage for existing folks?

mintapi/api.py
```diff
@@ -35,6 +36,10 @@
 except ImportError:
     pd = None
 
+logger = logging.getLogger('mintapi')
+logger.addHandler(logging.StreamHandler())
+logger.setLevel(logging.INFO)
```
mrooney: Nice logging changes, thanks!

mintapi/api.py
```diff
@@ -163,46 +159,92 @@ def get_email_code(imap_account, imap_password, imap_server, imap_folder, debug=
...
                    mfa_input_callback=None, wait_for_sync=True,
                    wait_for_sync_timeout=5 * 60,
                    session_path=None, imap_account=None, imap_password=None,
-                   imap_server=None, imap_folder="INBOX"):
+                   imap_server=None, imap_folder="INBOX",
+                   use_chromedriver_on_path=False):
     if headless and mfa_method is None:
-        warnings.warn("Using headless mode without specifying an MFA method"
-                      "is unlikely to lead to a successful login. Defaulting --mfa-method=sms")
+        logger.warn("Using headless mode without specifying an MFA method"
```
mrooney: Will logger.warn go to the terminal like `warnings`?

README.md
```diff
@@ -7,11 +7,9 @@ Ensure you have Python 2 or 3 and pip (`easy_install pip`) and then:
 
 ```shell
 pip install mintapi
-brew cask install chromedriver # or sudo apt-get install chromium-chromedriver on Ubuntu/Debian
 ```
 
-Note that chromedriver must be version 59+ if you want to use headless mode. If not installing via pip,
-make sure to install the `install_requires` dependencies from setup.py yourself.
+`mintapi` does it scraping by driving an instance of Chrome (or chromium) and navigating, as a human user would, thought the login flows. Once logged in, the API allows programatic access to various Mint REST APIs. Selenium/WebDriver is used to accomplish this, and specifically, ChromeDriver under the hood. `mintapi` will download the latest stable release of chromedriver, unless --use_chromedriver_on_path is given. **NOTE: You must have [Chrome](https://www.google.com/chrome/) or [Chromium](https://www.chromium.org/getting-involved/dev-channel/) installed, and on the `stable` track.**
```
mrooney: `thought the login flows`. I assume this is supposed to be throughout, but I think this sentence fragment can just be removed. It is clear as is :)

Is the requirement to have Chrome or Chromium installed yourself new or just being noted more clearly now?

mintapi/api.py
jprouty: You're right, this change will only update the defaults and test locations. However, the new tests will continue to work for existing int call sites, based on the [python rules for truthiness](https://docs.python.org/2/library/stdtypes.html#truth-value-testing).

For defining the param and showing users how to use it, Boolean is the correct primitive (either it's true/false), not an int.

mintapi/api.py
jprouty: Sure thing! I wanted to add new lines for the chromedriver downloading bit, but only wanted to do so if clients had the option to change the log output. Using the standard logging module allows clients to grab a reference to the 'mintapi' logger, and change the level or routing in their own code.

mintapi/api.py
jprouty: Yes; if you read up on warnings, the intent is to make it easy to filter out based on type of warning. However, mintapi did not use any of their structured warnings to make use of this feature. In this case, it's basically identical, and due to the log level being at INFO (warning is more severe than info), all of the .warn calls will go to stderr (and also INFOs). The [StreamHandler](https://docs.python.org/2/library/logging.handlers.html#logging.StreamHandler) sets this up.

README.md
jprouty: Revised that sentence for clarity. Good catch!

The requirement to have chrome/chromium installed (and in a standard location) is not new. The *new* requirement introduced by this change, and very intentionally, is that you must have an up-to-date version of chrome. This is so the chromedriver version can match the chrome version. This is for the best, as we don't want users staying behind on old versions of chrome when connecting to sensitive sites. For advanced users that _do_ want this capability, they can use the new path argument param to stay behind/ahead of stable. 

You could imagine someone adding additional support for the [remaining supported selenium drivers](https://selenium-python.readthedocs.io/getting-started.html#example-explained) (Firefox, IE/Edge, [Safari has their own](https://developer.apple.com/documentation/webkit/testing_with_webdriver_in_safari)). That, coupled with some crazy platform specific checks to detect which browsers are installed, and what the system default browser is, could be used to auto-detect the right driver to use. But that's a large burden on this module to support (seems like a good idea for a new python project, maybe?), especially since the DOM pieces could need special flows on a per browser X platform basis.

jprouty:
I've made the requested changes; I'll let you resolve the various threads when you're happy.

mintapi/api.py
mrooney: Thanks! What I mean is, if someone is currently calling `get_email_code` with `debug` specified and they upgrade mintapi, their code will create an exception, won't it? I'd prefer for upgrades to not break existing code except in pretty important cases.

I like the idea of making them False/True instead of 0/1, but would prefer that `debug` is kept in its position and a deprecation warning issued for a couple releases instead. Let me know if that makes sense or if I'm misunderstanding. Looks ready to merge after we figure that part out!

jprouty: Sorry; I misunderstood! You're absolutely right. I've added a deprecation warning with instructions on the new method.

jprouty:
Please take another look!

jprouty:
Hi Mike, a friendly ping on this PR. Thanks!

mrooney:
Thanks, I'll check this out as soon as I get a chance!

>

jprouty:
@mrooney I've merged upstream back into my PR. I also added another param to dictate where chromedriver should be downloaded. See the last commit for context as to why this is important/necessary in some cases.

mrooney:
Thanks for your patience! I like the changes and am reviewing it now. However, running into:

```
Traceback (most recent call last):
  File "mintapi/api.py", line 1412, in <module>
    main()
  File "mintapi/api.py", line 1322, in main
    chromedriver_download_path=options.chromedriver_download_path
  File "mintapi/api.py", line 438, in create
    return Mint(email, password, **opts)
  File "mintapi/api.py", line 434, in __init__
    chromedriver_download_path=chromedriver_download_path)
  File "mintapi/api.py", line 527, in login_and_get_token
    chromedriver_download_path=chromedriver_download_path)
  File "mintapi/api.py", line 273, in get_web_driver
    chromedriver_download_path))
  File "mintapi/api.py", line 220, in get_stable_chrome_driver
    local_executable_path)
  File "mintapi/api.py", line 193, in get_chrome_driver_major_version_from_executable
    return version_pattern.match(version.decode()).groupdict()['major']
AttributeError: 'NoneType' object has no attribute 'groupdict'
```

FWIW, `version.decode()` is `u'ChromeDriver 2.41.578706 (5f725d1b4f0a4acbf5259df887244095596231db)\n'` Let me know if I'm doing something wrong or if that seems like a valid error.

jprouty:
Great catch @mrooney. The issue is that you still have the old version of chromedriver hanging around (and so will others!), that uses version numbers prior to the new chrome version.major.minor.patch "system". This case is now properly handled (please verify). Thanks!

mrooney:
Thanks for the quick turnaround and I like the logging improvements! Now I'm getting a new one; is it possible to catch this (or detect versions that are too old) and provide relevant upgrade instructions?

```
Traceback (most recent call last):
  File "mintapi/api.py", line 1427, in <module>
    main()
  File "mintapi/api.py", line 1337, in main
    chromedriver_download_path=options.chromedriver_download_path
  File "mintapi/api.py", line 453, in create
    return Mint(email, password, **opts)
  File "mintapi/api.py", line 449, in __init__
    chromedriver_download_path=chromedriver_download_path)
  File "mintapi/api.py", line 542, in login_and_get_token
    chromedriver_download_path=chromedriver_download_path)
  File "mintapi/api.py", line 287, in get_web_driver
    chromedriver_download_path))
  File "/Users/rooney/Library/Python/2.7/lib/python/site-packages/selenium/webdriver/chrome/webdriver.py", line 81, in __init__
    desired_capabilities=desired_capabilities)
  File "/Users/rooney/Library/Python/2.7/lib/python/site-packages/selenium/webdriver/remote/webdriver.py", line 157, in __init__
    self.start_session(capabilities, browser_profile)
  File "/Users/rooney/Library/Python/2.7/lib/python/site-packages/selenium/webdriver/remote/webdriver.py", line 252, in start_session
    response = self.execute(Command.NEW_SESSION, parameters)
  File "/Users/rooney/Library/Python/2.7/lib/python/site-packages/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/Users/rooney/Library/Python/2.7/lib/python/site-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.SessionNotCreatedException: Message: session not created: This version of ChromeDriver only supports Chrome version 83
```

jprouty:
I thought about this and I haven't come up with a better solution. What it's telling you is that your version of chromedriver (which is now current/latest) is not compatible with your version of chrome or chromium (depends on your system, and what's installed). Specifically, it means your chrome is most likely out of date.

We could try to catch the exception, and suggest to the user that they update their browser for security sake. However, SessionNotCreatedException can be thrown for many reasons, and I'd hate to hide or obfuscate the underlying reason. We would have to string match for "This version of ChromeDriver ..." and rethrown/wrap the exception. WDYT?

The only other work around, which discussed above, would be to try to locate and determine the actual version of chrome (or chromium) someone has installed, and then find the most recent release of chromebox that matches/supports that version. This gets complicate right out of the gate, and I'd rather avoid this altogether, as it's more prudent that the user update their browser for their own sake (security/stability).

mrooney:
Yeah that's fair. When I hit this, the first thing I did was look in the README for a mention of this or instructions for upgrading, which I didn't notice. So I think an acceptable solution would be to just cover instructions for upgrading in the README with a mention of this exception, so if someone runs into it, they at least can find the right steps to fix it. What do you think?

jprouty:
Ha, sorry for the slow reply.

I do have a note in bold in the README. Do you want a specific mention of this exception? Specific upgrade instructions for chrome/chromium will vary by platform and seem a bit out of place for this project's README.

mrooney:
No worries! I like the text in the README, although it just mentions you need Chrome and Mint handles the rest, so reading that I felt unsure what the next step was. Maybe just add there that if you get an error about ChromeDriver version, to update your browser (Chrome / Chromium)? Then I think it is clear :)

jprouty:
How about this?

mrooney:
That looks great, thanks!!!

mrooney:
Thanks for all the work, merged and released to pypi as 1.43 :)

jprouty:
Thank you Mike for the review and maintaining the project!

As an FYI, I'll likely be removing some of the chromedriver logic in my app, and instead depending directly on this library so users of my code have one version/logic path for all the chromedriver logic/management.

PR #204, opened 2020-02-24
added --attention flag to get notification of any accounts that need …

jorythompson:
I added an --attention flag to to get any notification of accounts that need attention.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 build
 dist
 env
+chromedriver
 __pycache__
 *.pyc
 *.egg-info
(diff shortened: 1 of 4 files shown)


jorythompson:
I'm not really happy with the way I implemented this (having  global status_message in get_web_driver().  I wanted to return attention from get_web_driver() but the tests were giving me a hassle about the number of values returned.

Why is get_web_driver() not a member method?  This would alleviate my issue as well.

Hope this helps someone as much as it does me.

mrooney:
Thanks! So this is not something you think should be reviewed and merged at this time? Just informational?

jorythompson:
Well.... actually I was trying to merge it but I closed it instead.  I am
not really familiar with contributing to other's repos :-(
I was in process of duplicating it so I could submit it again.

Any suggestions?

thanks for your hard work on this project,
Jordan

jorythompson:
I deleted it and created a new one.

I'll get the hang of github eventually!

Jordan

PR #205, opened 2020-02-24
Added attention to get information about accounts that may need atten…

jorythompson:
Added an attention variable to indicate there may be some accounts that need attention.
Added chromedriver to .gitignore
Updated README to accommodate attention flag.  Changed wording on parameters to indicated username and password are required

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,13 +110,13 @@ Run it as a sub-process from your favorite language; `pip install mintapi` creat
                    [--budgets | --budget_hist] [--net-worth] [--extended-accounts] [--transactions]
                    [--extended-transactions] [--credit-score] [--credit-report] [--start-date [START_DATE]]
                    [--include-investment] [--skip-duplicates] [--show-pending]
-                   [--filename FILENAME] [--keyring] [--headless]
+                   [--filename FILENAME] [--keyring] [--headless] [--attention]
                    [--mfa-method {sms,email}]
-                   [email] [password]
+                   email password
 
     positional arguments:
-      email                 The e-mail address for your Mint.com account
-      password              The password for your Mint.com account
+      email                 The e-mail address for your Mint.com account (required)
+      password              The password for your Mint.com account (required)
 
     optional arguments:
       -h, --help            show this help message and exit
@@ -160,6 +160,7 @@ Run it as a sub-process from your favorite language; `pip install mintapi` creat
       --no_wait_for_sync    Do not wait for accounts to sync
       --wait_for_sync_timeout
                             Number of seconds to wait for sync (default is 300)
+      --attention.          Get notice if there are any accounts that need attention
 
     >>> mintapi --keyring email@example.com
     [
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -110,13 +110,13 @@ Run it as a sub-process from your favorite language; `pip install mintapi` creat
...
+                   [--filename FILENAME] [--keyring] [--headless] [--attention]
                    [--mfa-method {sms,email}]
-                   [email] [password]
+                   email password
 
     positional arguments:
-      email                 The e-mail address for your Mint.com account
-      password              The password for your Mint.com account
+      email                 The e-mail address for your Mint.com account (required)
+      password              The password for your Mint.com account (required)
```
mrooney: Password is actually not required, that's what `--keyring` is for :)

mintapi/api.py
```diff
@@ -173,12 +173,15 @@ def get_email_code(imap_account, imap_password, imap_server, imap_folder, debug=
...
 
+status_message = None
+
 
 def get_web_driver(email, password, headless=False, mfa_method=None,
                    mfa_input_callback=None, wait_for_sync=True,
                    wait_for_sync_timeout=5 * 60,
                    session_path=None, imap_account=None, imap_password=None,
                    imap_server=None, imap_folder="INBOX"):
+    global status_message
```
mrooney: I would love to see this done without the use of globals, even if you have to stick it as a property on the `driver` object that is returned.

jorythompson:
- I'll make the change to the README as you suggested.
   - I could add status_message to driver as you suggest, but it's not very
   intuitive.  IMHO it would be better to return status_message along with
   driver (see below) but the tests were failing and I don't know enough about
   nose and Travis to fix it. If you agree, point me in the right direction
   I'll make the changes there instead of just adding the message to driver:

so it becomes:

return driver, status_message


instead of:

driver.status_message = status_message
return driver


Let me know what you prefer.

On Tue, Feb 25, 2020 at 10:11 PM Mike Rooney <notifications@github.com>
wrote:
>
> @mrooney requested changes on this pull request.
>
> Thanks for contributing!
>
> ________________________________
>
> In README.md:
>
> >
>      positional arguments:
> -      email                 The e-mail address for your Mint.com account
> -      password              The password for your Mint.com account
> +      email                 The e-mail address for your Mint.com account
(required)
> +      password              The password for your Mint.com account
(required)
>
> Password is actually not required, that's what --keyring is for :)
>
> ________________________________
>
> In mintapi/api.py:
>
> >
>  def get_web_driver(email, password, headless=False, mfa_method=None,
>                     mfa_input_callback=None, wait_for_sync=True,
>                     wait_for_sync_timeout=5 * 60,
>                     session_path=None, imap_account=None,
imap_password=None,
>                     imap_server=None, imap_folder="INBOX"):
> +    global status_message
>
> I would love to see this done without the use of globals, even if you
have to stick it as a property on the driver object that is returned.

mintapi/api.py
jorythompson: I changed get_web_driver to return a tuple and I updated the test code to support the changes.

jorythompson:
I believe that I have incorporated your requests.  Please let me know if I missed anything.

PR #209, opened 2020-03-26
Suppress pandas future warning

dzg:


mrooney:
Thanks! Can you show what the normal warning is and explain why it makes sense to ignore it?

dzg:
> Thanks! Can you show what the normal warning is and explain why it makes sense to ignore it?

When I run my data grabber from shell, I get:
`/usr/local/lib/python3.7/site-packages/mintapi/api.py:694: FutureWarning: The pandas.np module is deprecated and will be removed from pandas in a future version. Import numpy directly instead`

I'm using it as part of bitbar, so it returns an error.

mrooney:
Is it not possible to just update the code to import numpy directly instead of hiding a valid warning? Just curious. Thanks for your efforts!

dzg:
> Is it not possible to just update the code to import numpy directly instead of hiding a valid warning? Just curious. Thanks for your efforts!

In my tests, `import numpy` did not help with the error message.

nelsonauner:
I am also pro updating code instead of hiding a valid warning. I've updated the code in https://github.com/mrooney/mintapi/pull/211

mrooney:
Thanks, will close in favor of #211, thanks for bringing this to my attention though, greatly appreciated :)

PR #211, opened 2020-04-11
Remove deprecated call to pd.np.nan

nelsonauner:
`pd.np.nan` is deprecated. You could import `np` and then use `np.nan` (which I'm fine doing, if you prefer), but pandas has added `pd.NA` which plays nicely with dataframes. See
https://stackoverflow.com/questions/60115806/pd-na-vs-np-nan-for-pandas
and
https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html#experimental-na-scalar-to-denote-missing-values

Original warning:
```
/lib/python3.7/site-packages/mintapi/api.py:692: FutureWarning: The pandas.np module is deprecated and will be removed from pandas in a future version. Import numpy directly instead
  .replace('uncategorized', pd.np.nan))
```

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -705,7 +705,7 @@ class Mint(object):
         df = pd.read_csv(s, parse_dates=['Date'])
         df.columns = [c.lower().replace(' ', '_') for c in df.columns]
         df.category = (df.category.str.lower()
-                       .replace('uncategorized', pd.np.nan))
+                       .replace('uncategorized', pd.NA))
         return df
 
     def populate_extended_account_detail(self, accounts):  # {{{


mrooney:
Thanks, great fix!

dzg:
This works, thank you!
Just wondering ... how/when does `pip` know about this new version? I had to manually sync from github to get this fix. (`pip ... --upgrade` didn't work)

nelsonauner:
@dzg The version of `mintapi` is tracked in https://github.com/mrooney/mintapi/blob/master/setup.py#L13.

I didn't increment the  version (in retrospect, I probably should have done a minor version bump) so pip does not view this fix as a new version, and upgrade will not do anything. 

As you can see the version is 4 months old despite active development. I'll defer to @mrooney  on when/how he wants to manage versioning. If there were instructions in say, a `CONTRIBUTING.md` file, I would follow those guidelines. But it's a great free service and I'm happy to help however

dzg:
OK, thanks, I don't mind manually upgrading.

mrooney:
Thanks the reminder, pushed to pypi as 1.42!

PR #220, opened 2020-07-26
allow close() method to clean up on login failure/exception

### LINKED ISSUE #190: After running, chromedriver leaves a bunch of zombies
I'm running this as a script every hour. But chromedriver leaves behind a bunch of idle processes after every run.

Any way to get chromedriver to exit cleanly?

Thanks!

mzakharo:
This CL separates creation of the selenium 'driver' from  performing login/token retrieval.  This allows for the API to be used in a long running process without creating zombie ```chromedriver``` and ```chrome``` instances if an exception occurs during execution of ```__init__()``` method. Potentially fixes #190 .  The new  usage pattern becomes the following:

```python
    mint = mintapi.Mint(headless=True, use_chromedriver_on_path=True)
    try:
        mint.login_and_get_token(email, password, mfa_method='sms')
        print(mint.get_budgets())
    finally:
        mint.close()

```

In my usecase, I am using ```mintapi``` in a long running python process, and very often,  instantiation of ``mintapi.Mint()``` class exits with the following exception:
```
Mint sync apparently incomplete after timeout. Data retrieved may not be current.
StaleElementReferenceException: Message: stale element reference: element is not attached to the page document
```

Without this change, during this exception, it is impossible to call ```close()``` method to clean-up selenium driver and retry without exiting the running interpreter session.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -255,36 +255,12 @@ def get_stable_chrome_driver(download_directory=os.getcwd()):
     return local_executable_path
 
 
-def get_web_driver(email, password, headless=False, mfa_method=None,
+def login(driver, email, password, mfa_method=None,
                    mfa_input_callback=None, wait_for_sync=True,
                    wait_for_sync_timeout=5 * 60,
-                   session_path=None, imap_account=None, imap_password=None,
-                   imap_server=None, imap_folder="INBOX",
-                   use_chromedriver_on_path=False,
-                   chromedriver_download_path=os.getcwd()):
-    if headless and mfa_method is None:
-        logger.warning("Using headless mode without specifying an MFA method"
-                       "is unlikely to lead to a successful login. Defaulting "
-                       "--mfa-method=sms")
-        mfa_method = "sms"
+                   imap_account=None, imap_password=None,
+                   imap_server=None, imap_folder="INBOX"):
 
-    chrome_options = ChromeOptions()
-    if headless:
-        chrome_options.add_argument('headless')
-        chrome_options.add_argument('no-sandbox')
-        chrome_options.add_argument('disable-dev-shm-usage')
-        chrome_options.add_argument('disable-gpu')
-        # chrome_options.add_argument("--window-size=1920x1080")
-    if session_path is not None:
-        chrome_options.add_argument("user-data-dir=%s" % session_path)
-
-    if use_chromedriver_on_path:
-        driver = Chrome(options=chrome_options)
-    else:
-        driver = Chrome(
-            options=chrome_options,
-            executable_path=get_stable_chrome_driver(
-                chromedriver_download_path))
     driver.get("https://www.mint.com")
     driver.implicitly_wait(20)  # seconds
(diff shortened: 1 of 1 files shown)


mintapi/api.py
```diff
@@ -514,32 +527,33 @@ def post(self, url, **kwargs):
...
 
     def login_and_get_token(self, email, password, mfa_method=None,
-                            mfa_input_callback=None, headless=False,
-                            session_path=None, imap_account=None,
+                            mfa_input_callback=None,
+                            imap_account=None,
                             imap_password=None,
                             imap_server=None,
-                            imap_folder=None,
+                            imap_folder='INBOX',
```
mrooney, line 534: Thanks for this great work! Is this INBOX change intended to be a part of this PR and could you describe the change if so?

mzakharo: This makes the change more consistent with the API it replaces. 'INBOX' was set as default before part of '__init__' function, but no longer, since __init__ is now only used to create selenium 'driver'.   See https://github.com/mrooney/mintapi/pull/220/files/c4bca605e950e21eaa58d84b734fa03c7afd9cea#diff-4117f091b0b8c6d18c2469a6f2452c55L433

mrooney:
Thanks for the updates! My only question/concern is if this will break usage for existing users because login_and_get_token is no longer in the __init__? I see you updated the README to mention calling this which is great, but I don't like to break upgrades unless it can't be avoided. Let me know if I'm just misunderstanding that change :)

mzakharo:
> Thanks for the updates! My only question/concern is if this will break usage for existing users because login_and_get_token is no longer in the **init**? I see you updated the README to mention calling this which is great, but I don't like to break upgrades unless it can't be avoided. Let me know if I'm just misunderstanding that change . You can fix the problem by duplicating ```close()``` logic during login while you do __init__(), but that just makes the code more complicated.   I dont see another solution, either more complicated code, or break the API. It is up to you I guess. 

> Thanks for the updates! My only question/concern is if this will break usage for existing users because login_and_get_token is no longer in the **init**? I see you updated the README to mention calling this which is great, but I don't like to break upgrades unless it can't be avoided. Let me know if I'm just misunderstanding that change :)

Unless you want to do try/finally and close() logic inside the __init__() call,, I dont see another option.

mrooney:
> Unless you want to do try/finally and close() logic inside the **init**() call,, I dont see another option.

Is there any compelling reason to avoid doing that? If that's the solution, why not? Breaking the usage for all existing users should be done only when there is no reasonable option. Thanks and let me know your thoughts :)

mzakharo:
> > Unless you want to do try/finally and close() logic inside the **init**() call,, I dont see another option.
> 
> Is there any compelling reason to avoid doing that? If that's the solution, why not? Breaking the usage for all existing users should be done only when there is no reasonable option. Thanks and let me know your thoughts :)

As you wish. Please take a look at the new CL.

jaron-l:
Closed in favor of #288

Issue #217, opened 2020-07-11
Error in atexit._run_exitfuncs:


j450h1:
I enter this in the terminal: ```mintapi --keyring --headless {you@example.com}```

I then enter my password when prompted and the correct MFA. Then I see a JSON response of my accounts, but at the very bottom this error appears. Is everything fine and this is just a red herring?

```
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/Users/jas/miniconda3/lib/python3.7/site-packages/mintapi/api.py", line 476, in close
    self.driver.find_element_by_id('link-logout').click()
  File "/Users/jas/miniconda3/lib/python3.7/site-packages/selenium/webdriver/remote/webelement.py", line 80, in click
    self._execute(Command.CLICK_ELEMENT)
  File "/Users/jas/miniconda3/lib/python3.7/site-packages/selenium/webdriver/remote/webelement.py", line 633, in _execute
    return self._parent.execute(command, params)
  File "/Users/jas/miniconda3/lib/python3.7/site-packages/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/Users/jas/miniconda3/lib/python3.7/site-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.ElementClickInterceptedException: Message: element click intercepted: Element <a id="link-logout" class="navbar-link" href="/logout.event?task=explicit&amp;country=">...</a> is not clickable at point (704, 32). Other element would receive the click: <div data-automation-id="ModalDialog--container" open="" role="dialog" class="StyledModalDialog__Container-cn89yw-1 iIlfZT" style="z-index: 102;">...</div>
  (Session info: headless chrome=83.0.4103.116)

```

mrooney:
Should be fixed in master after merging #221, thanks for the report!

Issue #215, opened 2020-06-03
`mint.close()` not working


dzg:
Recently I started getting this when doing `mint.close()`:

`selenium.common.exceptions.WebDriverException: Message: unknown error: Element <a id="link-logout" class="navbar-link" href="/logout.event?task=explicit&amp;country=">...</a> is not clickable at point (704, 32). Other element would receive the click: <div data-automation-id="ModalDialog--container" open="" role="dialog" class="StyledModalDialog__Container-cn89yw-1 iIlfZT" style="z-index: 102;">...</div>`

Full traceback:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/mintapi/api.py", line 400, in close
    self.driver.find_element_by_id('link-logout').click()
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/remote/webelement.py", line 80, in click
    self._execute(Command.CLICK_ELEMENT)
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/remote/webelement.py", line 633, in _execute
    return self._parent.execute(command, params)
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.WebDriverException: Message: unknown error: Element <a id="link-logout" class="navbar-link" href="/logout.event?task=explicit&amp;country=">...</a> is not clickable at point (704, 32). Other element would receive the click: <div data-automation-id="ModalDialog--container" open="" role="dialog" class="StyledModalDialog__Container-cn89yw-1 iIlfZT" style="z-index: 102;">...</div>
  (Session info: headless chrome=83.0.4103.61)
  (Driver info: chromedriver=2.41.578706 (5f725d1b4f0a4acbf5259df887244095596231db),platform=Mac OS X 10.14.6 x86_64)
```

robwil:
I believe this happens because Mint is now launching this modal on login:

![image](https://user-images.githubusercontent.com/605546/83952575-b699ed80-a807-11ea-870b-431eef8052f3.png)

Probably need to detect the modal and explicitly close it before trying to initiate the logout, but haven't had time to try that yet.

jaron-l:
[...]

PR #221, opened 2020-08-02
Remove logout step

jaron-l:
This pull request should solve #217 , #215, and maybe #190 depending on what that error actually is (I suspect it because it has the same symptom but really could be any error).

As you can see from the changes, all I did was remove the logout step as I believe it to be unnecessary and error prone. I tested on my laptop and could find no difference in functional workflow. When you close the browser, it automatically logs out anyways.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -471,12 +471,6 @@ class Mint(object):
         if not self.driver:
             return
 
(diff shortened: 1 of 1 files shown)


mrooney:
Hey thanks, great point!

PR #222, opened 2020-08-09
Remove extra blank line that fails flake8

jaron-l:
I noticed that the build is failing and it appears to me it is because of this one line that is causing flake8 to fail.


mrooney:
Thank you kindly!

PR #228, opened 2020-10-13
A fix and an enhancement to budgets

mcronce:
This makes two changes
* In `get_budgets()`, when not including history, uses the correct key for finding the most recent spending data - this may have only been a semantic fix, though, because (anecdotally speaking) I don't think I was seeing any actual issues that this would have caused previously
* Adds a `parent` field to budgets
  * Does so by changing `get_category_from_id()` to return an object instead of a string - are there any concerns here in terms of downstream breakage?  If so I can keep the API surface the same and add a separate function that returns the object for internal use.
  * I didn't do a `try`/`except` block or LBYL check when accessing the `parent` field on returned categories, because (again, anecdotally speaking) I didn't see any issues with my specific set of returned data.  I can always add them, though, if there's a concern.

Cheers :)

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -940,7 +940,7 @@ class Mint(object):
                     str(max(map(int, response['data']['income'].keys())))
                 ]['bu'],
                 'spend': response['data']['spending'][
-                    str(max(map(int, response['data']['income'].keys())))
+                    str(max(map(int, response['data']['spending'].keys())))
                 ]['bu']
             }
 
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -928,10 +928,9 @@ class Mint(object):
             for month in budgets.keys():
                 for direction in budgets[month]:
                     for budget in budgets[month][direction]:
-                        budget['cat'] = self.get_category_from_id(
-                            budget['cat'],
-                            categories
-                        )
+                        category = self.get_category_from_id(budget['cat'], categories)
+                        budget['cat'] = category['name']
+                        budget['parent'] = category['parent']['name']
 
         else:
             # Make the skeleton return structure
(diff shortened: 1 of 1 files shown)


mrooney:
Thanks for your contribution!

> are there any concerns here in terms of downstream breakage? If so I can keep the API surface the same and add a separate function that returns the object for internal use.

Yes, I know lots of folks use this for various things so I strongly prefer to avoid breaking APIs as I know how frustrating it is to upgrade and have your code break. I like the separate method approach. To avoid copy-pasting code, perhaps it is possible move the logic to your new function and then have the existing one simply call yours, get the object, and return the string?

mintapi/api.py
```diff
@@ -940,17 +939,16 @@ def mos_to_yrmo(mos_frm_zero):
                     str(max(map(int, response['data']['income'].keys())))
                 ]['bu'],
                 'spend': response['data']['spending'][
-                    str(max(map(int, response['data']['income'].keys())))
+                    str(max(map(int, response['data']['spending'].keys())))
```
mrooney, line 942: So you are saying you don't actually notice a bug with the current behavior, or a change in behavior when you change this? That's really confusing since it seems like it should be broken one way or the other...haha.

mcronce: Haha it sure does.  From looking at responses I get back, it seems like the keys in data.income and data.spending _just so happen_ to be the same.  I'm guessing because they're referring to timeframes?  (Months since epoch?)

mcronce:
> To avoid copy-pasting code, perhaps it is possible move the logic to your new function and then have the existing one simply call yours, get the object, and return the string?

That is exactly what I planned to do if your response was to retain the original functionality ;)

Just pushed up a change that does precisely that - let me know if there are any other concerns

mcronce:
OK, just pushed it up for realsies this time

mrooney (approved):
Looks great, thanks!

PR #233, opened 2020-12-05
Add a few things to make development easier

jaron-l:
Summary:
1. Added .vscode/ and chromedriver to .gitignore
2. Added mintapi to path of test.py
3. Added unittest.main()
4. Added requirements file for developers


mrooney:
Thanks, looks good! Do you have any idea though why the requirements.txt is showing up as a binary file with no diff available in the diff section?

jaron-l:
Ah, I did notice that but neglected to debug it. Looks like it's because it's in UTF-16 instead of UTF-8. I fixed it.

mrooney (approved):

mrooney:
Thanks! It does seem like now there's some duplication between dependencies in setup.py vs requirements.txt, but overall seems like a win. If you've got any thoughts to reduce that let me know!

jaron-l:
I agree that there's duplication that can be confusing. To be clear, I think putting the requirements.txt is mostly only helpful to developers because they want to be able to set up an environment to work on mintapi without installing the pypi version of mintapi as setup.py would. 

I don't know best practices but two opinions are shown [here](https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py). 

The second post has some good points about how requirements.txt is for developers of the package and needs to be specific for what has been tested (I just pinned versions FYI), while setup.py needs to be as loose as possible because it is unknown what environment it's going to. This seems to make the most sense to me.

PR #237, opened 2020-12-30
Fix chromedriver RE logic to prevent issue with greedy matching. Also fixes win10 error.

jprouty:
The leading wildcards in the version pattern with `.*` was causing the leading digit on the major version to be removed from the match. The logic around subprocess is modified to support running from a detached context, with no explicit stdin/out. This happens readily when running on win10 with a PyInstaller binary.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -176,8 +176,8 @@ CHROME_ZIP_TYPES = {
     'win64': 'win32'
 }
 version_pattern = re.compile(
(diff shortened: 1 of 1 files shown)


jprouty:
Hi @mrooney ! Let me know if you have any questions. Should be pretty safe/straightforward!

mrooney:
Thanks for this, looks great!

jprouty:
Thanks Mike for the quick merge and happy new year!

PR #240, opened 2021-01-04
Change method for getting current year for odate

KOEPKEJR:
Mint.com returns transactions from the current year as a string without the year (i.e. 'Jan 1').  The function json_date_to_datetime() attempts to resolve this shortcoming by appending the current year.  In certain scenarios at the start of a year the isocalendar function may return the previous year.  Updating to use date.today().year.


mrooney:
Thank you!

PR #242, opened 2021-01-16
Allows user to explicitly specify path to preferred [Chrome/Chromium] browser

llorracc:
Without further configuration, mintapi launches the system default version of Chrome.  That can be inconvenient if (for example) you are using chrome for other things.  This allows the user to specify a path to an alternative binary executable, which runs independently of whatever browsers the user might already have going. 

Example:

mintapi me@my.domain.org --alternative-browser-path /usr/local/share/chromium-extra-binary

The code right now is formatted to make it easy to see the parallels to the existing code for "chromedriver-download-path" rather than for elegance.  Should be reformatted to look better before merging.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -262,7 +262,7 @@ def get_stable_chrome_driver(download_directory=os.getcwd()):
     return local_executable_path
 
 
-def _create_web_driver_at_mint_com(headless=False, session_path=None, use_chromedriver_on_path=False, chromedriver_download_path=os.getcwd()):
+def _create_web_driver_at_mint_com(headless=False, session_path=None, use_chromedriver_on_path=False, chromedriver_download_path=os.getcwd(), alternative_browser_path=False):
     """
     Handles starting a web driver at mint.com
     """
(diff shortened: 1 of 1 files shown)


mrooney:
Thanks for your contribution! Have you tried using the existing use_chromedriver_on_path and/or headless options to solve this?

jaron-l:
@mrooney he's talking about which chrome binary executable to use, i.e. if you have more than one chrome version installed. This is different from which chromedriver to use.

llorracc:
@jaron-l is right.  I was trying to solve the problem that running mintapi with a cron job meant that if the job ran while I was using Chrome the Chrome instance for mintapi would have various annoying interactions with my ongoing Chrome session.  This PR solves that problem by allowing the user to tell mintapi to use, say, a browser that they never use for any other purpose (e.g., Chromium).  Hence no interference.

jaron-l:
Can you add this to the README.md so that it is easy for others to see that this is an option?

mrooney:
+1 for README! Thanks for the contribution. This might also help raspberry pi users :)

burkematthew:
@jaron-l @mrooney I propose that we close this one now that #330 is open with the necessary requested adjustments.

PR #252, opened 2021-03-18
add intuit account selection and password after mfa

jasonkissinger:


mintapi/api.py
```diff
@@ -339,7 +342,7 @@ def _sign_in(email, password, driver, mfa_method=None, mfa_token=None,
         except (NoSuchElementException, StaleElementReferenceException, ElementNotVisibleException):
             pass
 
-        driver.implicitly_wait(1)  # seconds
```
mrooney, line 342: Cool set of changes, I didn't even know this was a thing! What is the thought behind removing this wait?

jasonkissinger: i'm pretty sure driver.implicitly_wait() does not perform a wait itself, but is just a setting on the driver for how long it will waitFor an element to be found in the DOM before throwing a NotFound exception.  setting this to 1 just says that all subsequent find element calls on driver will try for up to 1s before throwing NotFound.

i'm not a python guy so bit out of my element here... just basing this off my experiences doing this in java/groovy with selenium.

jaikumarm:
tried this pull request and seems to have resolved my issue. 

I run my custom mintapi script on a ubuntu srv at home via cron (daily), it looked like it was failing with the following error for the last few days. 

```selenium.common.exceptions.NoSuchElementException: Message: no such element: Unable to locate element: {"method":"css selector","selector":"[id="ius-sign-in-mfa-password-collection-current-password"]"``` 

With this PR the above issue seems to have been resolved.

so +1 for this PR from my perspective.

File: mintapi/api.py [hunk 2]
```diff
@@ -349,42 +352,60 @@ def _sign_in(email, password, driver, mfa_method=None, mfa_token=None,
...
+
+        # account selection screen -- if there are multiple accounts, select one
+        try:
+            select_account = driver.find_element_by_id("ius-mfa-select-account-section")
+            if intuit_account is not None:
+                account_input = select_account.find_element_by_xpath(
+                    "//label/span[text()='{}']/../preceding-sibling::input".format(intuit_account))
+                account_input.click()
+            driver.find_element_by_id("ius-sign-in-mfa-select-account-continue-btn").click()
+            time.sleep(2)  # hack... what can we consistently waitfor?
```
mrooney, line 392: Is it possible to find an element to wait for here? This is going to be a bit brittle depending on internet connections, latency, etc.

mrooney (changes requested):
Looks really good, thanks for your time and contributions. I just would love to see if there is a way to avoid the sleep(2)!

mintapi/api.py
jasonkissinger: agree... changed it from a click() to a submit().  the latter waits for the button's form to complete.  verified that still works with my multiple accounts.

mrooney (approved):
Thanks, awesome PR!

areagu:
i'm not sure but it looks like master built failed...

jasonkissinger:
> i'm not sure but it looks like master built failed...

Yeah, noticed it's been failing for 5 months

https://travis-ci.org/github/mrooney/mintapi/builds/735886735

Likely why we haven't gotten the last couple of fixes pushed up for pip.

areagu:
ok got it - i should just get the repository directly from here then (instead of relying on pip install command)
thanks

mrooney:
The pip releases are all done manually, it isn't related to CI. This isn't
on pip yet as I'm on vacation and don't have access to a machine with those
credentials :) It should be up next week! In the meantime please do test
the latest from here.

Issue #259, opened 2021-05-22
Login changed... again


nemws1:
Looks like the mint.com login changed again - this time to a "2 phase" login, where you enter in your username, click submit, and then are prompted for a password (or to have it send you a code). At least, that's what I'm getting now.

hudcap:
I've got a funny feeling this was fixed in #252 (and maybe some other PRs) but hasn't been pushed to PyPI yet.
Can anyone confirm/deny?
I'm going to install manually and try when I get a chance.

nemws1:
Well, I went and launched Chrome and manually looked at the code in api.py and started working on some suggested changes - then I ran mintapi non-headless (I usually run it headless) and it worked / looked like the old login (with username & password asked at the same time).

Tried just now and saw the same - I'm getting a login page asking for both username and password.  However, when I go to mint.com in Chrome or FF on my own - I get just a username prompt, have to click "Submit" and *THEN* I get a password prompt.

Not sure what's going on with 'mint.com' as to why I get a different login experience.  Guessing they're in the middle of migrating to a new login experience, though.

hudcap:
@nemws1 Do you use `session_path`? If not, that might be why you're getting the old login with `mintapi`.

I'm getting the following on both (username/email crossed out):
![image](https://user-images.githubusercontent.com/4990645/119423973-88404700-bcd2-11eb-9e8c-5754d1f3aceb.png)

nemws1:
Yes, I am using `session_path` - was getting inconsistent results without it.  However, I also attempted a "fresh" session (renamed by session directory) and I saw the same thing when I did - separate username (submit) then password (submit) process.  I've never had mint.com try to remember my username.

mrooney:
Hi there, #260 looks pretty simple, can anyone confirm or deny of that helps them out? PS, at the request of some users we've now got a Discord for debugging / chatting etc, in the README. Feel free to check it out since it can be hard to coordinate efficiently over GH issues sometimes :)

mrooney:
I just pushed that to pypi as 1.46 :)

PR #260, opened 2021-05-25
handle saved-account login page

hudcap:


hudcap:
I think this fixes #259 
It works for me when the login looks like this:
image

mrooney:
Thanks, looks solid to me!

Issue #253, opened 2021-03-24
Getting string indices must be integers error when calling mint.get_budgets in Python? 


jcsacto:
Greetings all, so I finally got past all the issues with the email only login prompt, but now I'm trying to do a simple call to mint.get_budgets() and getting the 'string indices must be integers' error.   Do we know of any issues going on with the budgets area of code by chance?  Please be advised, I am novice Python developer. : )

Thanks! 

Python: 
budgets = mint.get_budgets()

TypeError                                 Traceback (most recent call last)
<ipython-input-9-ab9c26f7088d> in <module>
      1 #Get this months budget data
----> 2 budgets = mint.get_budgets()
      3 

~\.conda\envs\mintEnv\lib\site-packages\mintapi\api.py in get_budgets(self, hist)
   1015                     category = self.get_category_object_from_id(budget['cat'], categories)
   1016                     budget['cat'] = category['name']
-> 1017                     budget['parent'] = category['parent']['name']
   1018 
   1019         return budgets

TypeError: string indices must be integers

jaron-l:
@mrooney this can probably be closed with #261 being already merged.

mrooney:
Thanks, I'll mark as a duplicate, and Jon I hope you are able to resolve with the latest version.

mrooney:
Duplicate of #261

jcsacto:
@mrooney  Thanks, unfortunately I'm now getting that error when trying to pull budget history, any advice? 
 mint.get_budgets(hist=2)

mrooney:
@jcsacto thanks for the update! Are you getting that with the latest mintapi, 1.53? If so, feel free to open a new report with our beautiful new bug template :D

PR #261, opened 2021-06-05
fix: Fix for 'string indices must be integers'

jaron-l:
The 'Uncategorized' budget's parent is a string not a dict.
Probably only affects those who put a specific limit on uncategorized.

Refs #253


mrooney:
LGTM, thanks, I'll release this to pypi!

PR #270, opened 2021-07-26
Do not attempt to send the password in the MFA phase unless the element is visible (and therefore interactable).

jprouty:
This fixes issues https://github.com/mrooney/mintapi/issues/265 and https://github.com/jprouty/mint-amazon-tagger/issues/76.

Also removes a stale comment (I just verified that the mfa path is correct and still functions correctly too.


mrooney:
Awesome, thank you!

jprouty:
Thanks for the quick review!

PR #286, opened 2021-09-19
chore: Add stale.yml configuration file for stale bot

### LINKED ISSUE #285: Repo Management: Integrate Stale Probot Into MintApi
In the past, when trying to help people to debug their issues, I've run across the problem where users won't respond for a request for more information on their issue. I think that issues that we've responded to and asked for more information should be closed after a period of time if we receive no response.

My proposition:
Whenever a maintainer of mintapi asks for more details on a issue, he should label the issue with 'awaiting-feedback'. If the issue does not receive more comments after 60 days, then the issue is closed. We would enforce this rule with the [stale probot app](https://github.com/marketplace/stale). I would be happy to set it up if @mrooney thinks this is a good idea.

jaron-l:
* This configuration will mark 'awaiting-feedback' labeled issues or pull requests with a stale label
if no one responds to the issue/PR within 60 days. It will then close it after 7 after the label if no futher activity results.

Fixes #285

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,13 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: ['awaiting-feedback']
+
+# Label to use when marking as stale
+staleLabel: stale
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not recieved 
+  an answer to a collaborator's question. It will be closed if no further activity
+  occurs. Thank you for your contributions.


mrooney:
Looks great, thanks for adding this! So the 60 days and 7 days are built-in to the bot, and we just configure the labels? And, the only manual part is then adding the `awaiting-feedback` label after asking a question?

jaron-l:
Yes, the 60 and 7 days are the defaults. I added the option to only look as awaiting-feedback because I thought that it would be better to only have it mark issues that we were waiting on, rather than no one has gotten around to fixing. And yes, the only manual part is adding the label awaiting-feedback. I couldn't think of a way to easily automate that part.

jaron-l:
If you want to see the other defaults, you can view them [here](https://github.com/probot/stale#:~:text=GitHub%20%2D%20probot%2Fstale%3A%20A,after%20a%20period%20of%20inactivity.)

mrooney (approved):
Looks great to me!

PR #288, opened 2021-09-20
fix: :bug: Fix but where mintapi leaves chrome instances if there is an error in sign in

### LINKED ISSUE #190: After running, chromedriver leaves a bunch of zombies

jaron-l:
* Wraps sign in and wait-for-sync with a try-except and close driver on exception

Fixes #190

This is a re-write of the stale pull request #220. I thought that that pull request had too many refactors built into the code and was too far behind the main branch that the pull request became confusing. This PR should be simpler but still have the core fix.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -629,7 +629,8 @@ class Mint(object):
             wait_for_sync_timeout=wait_for_sync_timeout,
             use_chromedriver_on_path=use_chromedriver_on_path,
             chromedriver_download_path=chromedriver_download_path)
-        self.token = self.get_token()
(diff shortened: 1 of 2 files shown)


jaron-l:
Planned work:

- [ ] Test that breaking sign in is handled without creating unresolved daemon processes

mrooney (approved):
Seems like a clean and simple approach, thanks for taking this up! I'll mark as approved pending the testing that you've mentioned :)

PR #292, opened 2021-09-23
update github urls to new location

mrooney:
This will point pypi folks at the right place as well as the badges :)


mrooney:
Fixes #293

PR #296, opened 2021-09-24
require selenium-requests>=1.3.3 in setup.py for pip

mrooney:
This should ensure you get the latest selenium-requests when pip installing as well: https://packaging.python.org/discussions/install-requires-vs-requirements/.


mrooney:
@jaron-l just a heads up, looks like this will help when installed directly via pip :)

PR #299, opened 2021-09-25
update stablebot to consider PRs stale after 31 days

mrooney:
After thinking about more, 60 days feels like a very long time to keep an issue open if we are waiting for feedback. 31 days + the 7 additional days after the comment feels quite generous for them or someone else to respond, and of course they can always respond later and we can open it back up. What do you think?


jaron-l (approved):
Sounds good to me!

Issue #278, opened 2021-08-30
Get Account data based on date.


timdau:
https://mint.intuit.com/trend.event Allows filtering based on date.  I would like to be able to get the account data based on a specified date.

searchQuery: {"reportType":"AA","chartType":"P","comparison":"","matchAny":true,"terms":[],"accounts":{"groupIds":["AA"],"accountIds":[],"count":25},"dateRange":{"period":{"label":"This month","value":"TM"},"start":"8/1/2021","end":"8/29/2021"},"drilldown":null,"categoryTypeFilter":"all"}

mrooney:
Have you tried the `--start-date` parameter, and does that accomplish what you are looking for? Or are you looking for an end date as well?

timdau:
I am looking for the end date.  I tried `--start-date` and it doesn't change anything in the output of --accounts.

timdau:
I am not sure if it helps but I would be happy to provide $200USD as a thank you gift to anyone who can provide this functionality in such a way that it gets merged into this repo.

jaron-l:
duplicate of #208

burkematthew:
Unless someone beats me to it, once I finish getting everything going with mintapi, I'll take a crack at this.

burkematthew:
I'd just like to get some clarification here on what the expectation is when filtering accounts based on date.  Is the expectation that the results are filtered based on the accounts opened during that date, the accounts that have activity during that date, etc?

burkematthew:
@timdau thanks for the clarification.

burkematthew:
So I went in and reviewed this a little bit more.  The section of Mint that you're referencing is really looking at trends and uses a different API call than the current handling of the account data.  

Presently, we use `bundledServiceController.xevent` to get the account data, while the trend section uses `trendData.xevent` with a different set of input parameters.  It could be something we explore, but with as major of a change as I'd imagine that would be, I'd want to get @mrooney and @jaron-l to give thoughts on this first.

burkematthew:
@jaron-l agreed.  I'll add this to my list of tasks.  Thanks!

burkematthew:
Noting on here that I had started to work on this and ran into some roadblocks and ended up pivoting.  Will revisit when I get a chance.

timdau:
Thanks @burkematthew, Happy New Year

timdau:
@burkematthew Who do I owe the $200USD as the thank you gift?

burkematthew:
[...]

Issue #208, opened 2020-03-22
Add an end_date Parameter to get_transactions Methods


jones-chris:
This library is very helpful!  

It would be great to be able to specify an end date parameter, like the `start_date` parameter, in the `get_transactions` method.

I can currently apply the end date by filtering the data frame for transactions more recent than the end date, but it would be great if `get_transactions` could add an end date to the mint HTTP query parameters as is discussed in this mint article:

https://help.mint.com/Accounts-and-Transactions/972190281/How-can-I-view-transactions-within-a-specific-date-range.htm

Thanks again for creating this!

mrooney:
Certainly, PRs welcome!

jones-chris:
I'll see if I can find some time 😄 !  This library is great 😄 !

PR #301, opened 2021-09-29
Adding support for an end-date argument.

burkematthew:
Addresses #278 and #208.  This is my first contribution to a project like this, so feedback is certainly welcome.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -58,6 +58,7 @@ def json_date_to_datetime(dateraw):
         newdate = convert_mmddyy_to_datetime(dateraw)
     return newdate
 
+
 def convert_mmddyy_to_datetime(date):
     try:
         newdate = datetime.strptime(date, '%m/%d/%y')
@@ -65,11 +66,13 @@ def convert_mmddyy_to_datetime(date):
         newdate = None
     return newdate
 
+
 def convert_date_to_string(date):
     date_string = None
     if date:
         date_string = date.strftime('%m/%d/%Y')
     return date_string
+
 
 def reverse_credit_amount(row):
     amount = float(row['amount'][1:].replace(',', ''))
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -55,9 +55,21 @@ def json_date_to_datetime(dateraw):
     try:
         newdate = datetime.strptime(dateraw + str(cy), '%b %d%Y')
     except ValueError:
-        newdate = datetime.strptime(dateraw, '%m/%d/%y')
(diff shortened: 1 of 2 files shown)


burkematthew:
Addresses #278 and #208.  This is my first contribution to a project like this, so feedback is certainly welcome.

mintapi/api.py
```diff
@@ -789,29 +799,22 @@ def get_transactions_json(self, include_investment=False,
             txns = data['set'][0].get('data', [])
             if not txns:
                 break
-            if start_date:
```
mrooney, line 792: Is the idea here that we can use startDate and endDate directly in the calls above, and don't need to filter ourselves anymore?

mintapi/api.py
```diff
@@ -760,6 +761,13 @@ def get_transactions_json(self, include_investment=False,
             start_date = datetime.strptime(start_date, '%m/%d/%y')
         except (TypeError, ValueError):
             start_date = None
+
+        # Converts the end date into datetime format - must be mm/dd/yy
+        try:
+            end_date = datetime.strptime(end_date, '%m/%d/%y')
+        except (TypeError, ValueError):
+            end_date = None
+
```
mrooney, line 770: Makes sense! My only thought here is that this is basically copy and pasted from above. Can this be abstracted into a little helper method and then you can simply call it twice with different dates?

mrooney (changes requested):
Thanks for your awesome contribution! Looks great overall, just a minor question / feedback or two.

mintapi/api.py
burkematthew: That's correct.  I was running into some friction with the fact that we were using the filtering to also break out of the loop.  So I couldn't exactly do the same thing that we did with start_date.  I explored a little more of the methods and saw that there was another instance of using startDate and endDate.  I then went ahead and tested out the behavior myself and got the desired results.

mintapi/api.py
burkematthew: Yes, definitely can!  I will do that.

mrooney (approved):
Looks awesome, thanks!

mrooney (changes requested):
Actually, I realized I had to approve the workflow checks for you as a first-time contributor, and it did hit very minor flake8 line spacings to fix: https://github.com/mintapi/mintapi/pull/301/checks?check_run_id=3810069327.

You should be able to run that locally and add the extra spaces, or if not, just add them on the lines mentioned :) Thanks again and looking forward to merging and releasing this!

PR #302, opened 2021-10-02
Extract argparser commands into array and add argument via loop

burkematthew:
Note that this will produce a merge conflict with #301 due to the addition of the end_date command line argument.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -169,7 +169,7 @@ ARGUMENTS = [
    (('password' , ), {'nargs': '?', 'default': None, 'help': 'The password for your Mint.com account'}),
    (('--accounts', ), {'action': 'store_true', 'dest': 'accounts', 'default': False, 'help': 'Retrieve account information (default if nothing else is specified)'}),
    (('--attention', ), {'action': 'store_true', 'help': 'Display accounts that need attention (None if none).'}),
-   (('--budgets', ), {'action': 'store_true', 'dest': 'budgets', 'default': None, 'help': 'Retrieve budget information'}),
+   (('--budgets', ), {'action': 'store_true', 'dest': 'budgets', 'default': False, 'help': 'Retrieve budget information'}),
    (('--budget_hist', ), {'action': 'store_true', 'dest': 'budget_hist', 'default': None, 'help': 'Retrieve 12-month budget history information'}),
    (('--chromedriver-download-path', ), {'default': os.getcwd(), 'help': 'The directory to download chromedrive to.'}),
    (('--credit-report', ), {'action': 'store_true', 'dest': 'credit_report', 'default': False, 'help': 'Retrieve full credit report'}),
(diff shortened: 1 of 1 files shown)


jaron-l:
What value does this add? I think this just adds a little more confusion as the developer now needs to identify the key values to how they are used in the argparse add_argument function. I think the current method is clearer.

burkematthew:
I find it redundant to have the same function copied over and over again, the difference of which is simply the input values.  I think this provides more scalability without a continued copy-and-paste effort.  

An example, too, as a follow-up to this change is I'm looking switching from argparse to configargparse, which is essentially the same library, except it also allows pulling input from a config file.  I need to do more work on that, but in making such a change, we'd only have to change the one instance in the loop for adding arguments to the parser.  

It doesn't add any additional work than is currently necessary.  Each line in the ARGUMENTS array only has the options relevant to its purpose, just like now.  

What makes it more confusing from your perspective?

jaron-l:
I think it gets more confusing because it separates the inputs from how they are used. Just looking at the `ARGUMENTS` is not descriptive enough to understand what it is doing and how it works. You have to go down to the loop (which is 1000 lines later but even if it was right after, my argument still stands) to figure out what it is and trace in your mind how the variable will be deconstructed. I don't think the copy-paste effort argument is a valid one because code should be [explicit](https://www.python.org/dev/peps/pep-0020/#id2). If it doesn't affect performance, then it should be as easy to understand as possible, even if it is more spelled out.

Don't mistake my opinions for not being grateful for your contributions. I think as we talk things out, we'll come to the best decision.

burkematthew:
No concern at all.  I'm happy to be able to participate in development and conversations like this.  At the end of the day, we all just want is the most beneficial for the application.  

I do see that there are some aspects of the change that simply result in more questions than answers.  Really, the criticism appears to be related to the fact that it's not very clear on what's going on.  A couple thoughts on ways that I could improve clarity here:

1.  At a minimum, the ARGUMENTS array can be moved closer to the implementation.
2. I could re-format how the ARGUMENTS array is written to make it easier to read\understand.  I do agree that deciphering what each line of the array is\does is more confusing than the original approach.
3. I could add some more comments surrounding the loop to give the full picture of what it's doing and what is needed to add a CL argument.
4. I could also look at reworking the for loop to not use the shorthand format, bringing more clarity to it as well.  

Thoughts on those?  If you think it might make you feel better about the new approach, then I can go ahead and make changes related to 1-4 and then we can discuss again.  If you still think with those changes that it won't be worth it, then we can go ahead and just move on and I can work on other features\changes.

mrooney:
> Gah... I still feel like the changes wouldn't be worth it. If something needs more documentation to explain how it works versus another way that is self-explanatory, I would always go with the self-explanatory. Keep it simple. If you want to "appeal", you can get @mrooney's input but I think my opinion is unchanged but would be happy to concede if I'm out voted.
> 

While I'm not sure what a perfect solution is, I do find the new code more organized and maintainable than the previous code. I think it is a net improvement. I agree it is too far away from where it is used, so if you move the ARGUMENTS definition inside the main() function where it is used (and fix the flake8 issues from the checks), I'm 👍 .

mrooney (approved):

PR #305, opened 2021-10-07
add 3.10 python for github build checks

mrooney:
It seems like this is the current version of Python, so we should add 3.10. We can remove them as they hit EOL, such as 3.6 on Dec 23.


jaron-l:
This error might be because we're using nose which has been deprecated for several years now. I propose we use a different test suite. Pytest is usually the most popular. Do you want me to try it?

PR #309, opened 2021-10-10
Switching from argparse to configargparse and supporting config file …

burkematthew:
…usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ an MFA prompt, you'll be prompted on the command line for your code, which by de
 goes to SMS unless you specify `--mfa-method=email`. This will also persist a browser
 session in $HOME/.mintapi/session to avoid an MFA in the future, unless you specify `--session-path=None`.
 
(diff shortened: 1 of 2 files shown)


jaron-l:
Could you add a test to show that it works?

burkematthew:
Yes.  I will work on that.

tests/test_driver.py
```diff
@@ -110,6 +111,14 @@ def test_when_sign_in_fails_then_logs_exception(self, mock_sign_in, mock_logger,
         mintapi.Mint('test', 'test')
         mock_logger.exception.assert_called_with(test_exception)
 
+    def test_config_file(self):
+        # verify parsing from config file
+        config_file = tempfile.NamedTemporaryFile(mode="wt", dir="/tmp", delete=False)
+        config_file.write("extended-transactions")
+        mint = mintapi.Mint()
+        arguments = mint.parse_arguments(["-c " + config_file.name])
+        self.assertEqual(arguments.extended_transactions, 'true')
```
jaron-l, line 120: The problem is not that the file doesn't exist, but the way that you were passing the arguments as a single item in the list. In addition, writes don't happen until you close or flush so I added the flush. Also, probably shouldn't specify the directory as your method wasn't cross-platform. Anyways, I tested the following and it works.
```suggestion
        config_file = tempfile.NamedTemporaryFile(mode="wt")
        config_file.write("extended-transactions")
        config_file.flush()
        mint = mintapi.Mint()
        arguments = mint.parse_arguments(["-c", config_file.name])
        self.assertEqual(arguments.extended_transactions, True)
        config_file.close()
```

jaron-l (changes requested):

Issue #60, opened 2015-12-21
Return parent and sub-categories?


felciano:
Right now if a subcategory (e.g. "Office Supplies") shows up under two parent categories in mint (e.g. "Business Expenses" and "Home Office") then they are indistinguishable in the exported CSV and JSON files. Would it be possible to return both parent and sub-categories for Mint transactions, for example using a category identifier?

mrooney:
That seems reasonable! Should we use representations like "Home Office/Office Supplies", or are you suggesting adding a field that contains a category ID, if such a thing exists?

felciano:
I think a `full_category_path` field using slash conventions is fine to start with, provided slashes in the original name are escaped.

A future version could actually structure this out by including both name and id for both parent and subcategory, but even then I expect it would be helpful to have a `full_category_path`. So I would vote for getting something simple out initially and then iterating if/as needed.

mrooney:
Sounds good, thanks! It isn't a feature I quite have time to add at the moment, though I'm happy to review/merge/release a PR if you find the time!

jaron-l:
I think if category nesting can only go one layer deep then having the separate columns would be easiest as no parsing would be required. I was actually not aware that sub-categories could have more than one parent so I don't know if you have have sub-sub-categories.

felciano:
Agree (not sure why I recommended the `full_category_path` above -- keeping them distinct seems cleaner)

PR #310, opened 2021-10-11
Add columns for parent category to the output of extended-transactions.

burkematthew:
* Saving changes to adding parent category

* If parent is Root, then do not populate the parent columns


burkematthew:
Enhancement as per requested on #60 .

jaron-l:
Can you add a test that verifies this?

mrooney:
Thanks!

PR #316, opened 2021-10-17
Change Account Continue Button to find by CSS Selector

egwoo:
At least on my account, finding by `id` doesn't work. On inspection, the Continue button has no `id` property, but it does have `data-testid="SelectAccountContinueButton"`. This patch updates the Account Selection method to work with both the current approach and on my account's approach.


jaron-l:
The implicit wait is probably not what you're looking for. See the [selenium documentation about waits](https://www.selenium.dev/documentation/webdriver/waits/). You might be looking for a normal sleep but I'm not sure why that's there.

egwoo:
Yeah you're right. In my initial testing, it didn't work until I added the wait. I just removed it and tested it again and it works now, so that line is unnecessary.

PR #330, opened 2021-10-30
Allow users to explicitly specify path to preferred Chrome browser.

burkematthew:
Primary credit to the original author of #242.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,11 @@ If you wish to simplify the number of arguments passed in the command line, you 
 
 ### Linux Distributions (including Raspberry Pi OS)
 
-If you're running mintapi in a server environment on an automatic schedule, consider running mintapi in headless mode if you don't need to see the login workflow. In addition, you'll want to use your distribution's package manager to install chromium and chromedriver. Make sure your distribution is up-to-date and then install/update Chromium (debian-family example): `apt install chromium-browser chromium-chromedriver`. Then use the option `use_chromedriver_on_path` either through the CLI or the python api so that mintapi doesn't try to find a matching chromedriver.
+If you're running mintapi in a server environment on an automatic schedule, consider running mintapi in headless mode if you don't need to see the login workflow. In addition, you'll want to use your distribution's package manager to install chromium and chromedriver. Make sure your distribution is up-to-date and then install/update Chromium (debian-family example): `apt install chromium-browser chromium-chromedriver`. Then use the option `use-chromedriver-on-path` either through the CLI or the python api so that mintapi doesn't try to find a matching chromedriver.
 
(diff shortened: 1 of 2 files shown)


mrooney:
Nice, thanks! Just for my understanding, what does this solve that can't be done by adding a PATH env var to the cron job invocation?

burkematthew:
@mrooney my understanding is that the user could be already using the binary executable in the PATH; but by using this command-line option, you get the flexibility to choose another executable separate from your primary workspace.

mrooney:
> @mrooney my understanding is that the user could be already using the binary executable in the PATH; but by using this command-line option, you get the flexibility to choose another executable separate from your primary workspace.

Yeah, I wasn't thinking of the user's PATH, but that you can specify PATH in the crontab in a couple different ways (at the top or as an env var for a specific task). Unless there's some reason that doesn't work that I'm missing, I'd prefer we just document how to use crontab better than add in extra code that duplicates a crontab feature. But please let me know if I'm missing/understanding something :)

burkematthew:
@mrooney honestly, I have zero preference. I simply took a stale PR and tried to get it moving, which seemed to be just to add content to the README. See the referenced PR above. Whatever y'all want to do is fine with me.

mrooney:
> @mrooney honestly, I have zero preference. I simply took a stale PR and tried to get it moving, which seemed to be just to add content to the README. See the referenced PR above. Whatever y'all want to do is fine with me.

Curious if @jaron-l has any preferences, but if this can be solved by simple prepending the command in cron or elsewhere with a `PATH` var, that seems better to leverage native OS features than add code here. But I don't have a strong opinion and don't mind being outvoted if this flag seems like a helpful idea. It just seems like it will need to be documented either way, so we could just as easily document using PATH. 🤷

jaron-l:
I can definitely see both sides here as well. I understand avoiding feature bloat that already has a workaround that could be just documented. One argument for this PR would be that selenium provides a way to do it in it's API and we are hiding that ability.

I hate to be that guy that complicates the discussion... but it almost feels like we could just make it easier by documenting how to start your own webdriver and pass that into mintapi to be used. Then any supported selenium API would be usable from the python API and would also be more future proof. For example, we would refactor _sign_in from get_web_driver and document how the user can pass in a set up driver and bypass _create_web_driver_at_mint_com. I could work on this PR request if the idea is hard to imagine.

burkematthew:
@jaron-l that's definitely a very intriguing idea and sounds like it is the best long-term solution, at least based on what I can tell from your description. I'd be fine with going that route or at least seeing how that'd look.

jaron-l:
> @jaron-l that's definitely a very intriguing idea and sounds like it is the best long-term solution, at least based on what I can tell from your description. I'd be fine with going that route or at least seeing how that'd look.

Take a look at #336. That's my initial draft of what I'm talking about. I documented how to use it in the README.md but you could use that same idea for any seleniumrequest driver.

jaron-l:
I'm going to close this in favor of #336 as it seems we are in agreement about going that route.

PR #331, opened 2021-10-30
Chore: Consolidating logic to build bundledServiceController requests and making POST requests

burkematthew:
This PR is just adding a little more DRYness to the handling of POST requests and certain URLs.

@mrooney @jaron-l let me know if you have any issues with this.

### DIFF
--- a/tests/test_driver.py
+++ b/tests/test_driver.py
@@ -28,6 +28,10 @@ try:  # read test_args file if supplied
 except (FileNotFoundError, AssertionError):
     test_args = None
 
+MINT_ROOT_URL = 'https://mint.intuit.com'
+MINT_ACCOUNTS_URL = 'https://accounts.intuit.com'
+MINT_CREDIT_URL = 'https://credit.finance.intuit.com'
+
 accounts_example = [{
     "accountName": "Chase Checking",
     "lastUpdated": 1401201492000,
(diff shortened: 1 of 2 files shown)


mrooney (changes requested):
Nice, great cleanup! Although it is minor, can we add a simple test for "build_bundledServiceController_url"?

tests/test_driver.py
```diff
@@ -28,6 +28,10 @@
 except (FileNotFoundError, AssertionError):
     test_args = None
 
+MINT_ROOT_URL = 'https://mint.intuit.com'
+MINT_ACCOUNTS_URL = 'https://accounts.intuit.com'
+MINT_CREDIT_URL = 'https://credit.finance.intuit.com'
```
mrooney, line 33: Can we import these from `api`, or is there a good reason to have them here as well? Just trying to avoid any duplication as we remove some others :)

mrooney (changes requested):

tests/test_driver.py
burkematthew: I was trying to build a constants file and pull from there and when that got sticky, I just went to hard code. I didn't even try to reference from API. I'll try that and see!

burkematthew: @mrooney there you go, should be better now!

mrooney (approved):

mrooney:
Nice, thanks!

PR #333, opened 2021-11-05
ci: Black linter

jaron-l:


jaron-l:
Noting here that it looks like black did break some of the code. For instance, Mint::get_transactions had it's return statement removed.

mrooney:
> Noting here that it looks like black did break some of the code. For instance, Mint::get_transactions had it's return statement removed.

That's no good! I tried to find this in the diff / branch commits and don't see it. Was it since changed and squashed / force pushed?

jaron-l:
I edited the original comment in the issue. It was my mistake. When I
rebased onto main, the deleted part of the merge conflict. Black was able
to regenerate successfully on a clean branch so I force pushed the
erroneous commit.

On Fri, Nov 5, 2021, 11:43 PM Michael Rooney 'notifications at github.com'
<jaron-github+reply+ADGNKUFKXAUOVS43IGIXBJF7SH3QBEVBNHHD4VPHG4=

mrooney:
@jaron-l okay, so to clarify my understanding, are you saying that Black still did remove a return statement, or that was just an error in the rebase? If the latter, it could still make sense to use it. As you said also, this is the one time it would be going over existing code, and in future cases it would be new changes / code where that type of thing would be hard to miss.

jaron-l:
I'm saying that black did *not* remove the return statement. I made the error when rebasing the branch onto main and resolving the merge conflicts. To prove this, I reran black on the head of main and showed that all the tests still ran. I forced push because I swapped out the broken test commit with the commit based on main because I wanted black to be run from the "clean" (unaltered by black) head of main.

I briefly skimmed the resulting changes from black and couldn't find anything beyond whitespace and quote mark changes.

.github/workflows/ci.yml
```diff
@@ -29,10 +29,9 @@ jobs:
       - name: Install dependencies
         run:
           python -m pip install --upgrade pip
-          pip install -r requirements.txt flake8 pytest
-      - name: Lint with flake8
-        run:
-          flake8 --ignore=E501 mintapi # pep8 flake ignoring long lines
+          pip install -r requirements.txt pytest
+      - name: lint
+        uses: psf/black@stable
```
mrooney, line 34: Awesome! So will this just fail if the code isn't properly styled, or does it actually make changes to the PR?

jaron-l: It just fails currently. It doesn't look like the support committing the changes made in the pipeline currently.

jaron-l:
@mrooney FYI, this PR will need a rebase every time we merge something into main before it due to it changing the CI tests and the high likelihood that formatting errors will be in the commits before it. Whenever you feel comfortable, you can approve it, but if it's alright, let me do the merge because I'll want to rebase and rerun black before actually merging.

mrooney:
@jaron-l makes sense, I guess if it isn't auto-committing, we should at least add a step that checks that the code *is* in sync w/ black, and add a note to README/Contributing on how to do this. I just want to make sure I understand our strategy going forward, as I'm excited about this change!

jaron-l:
> @jaron-l makes sense, I guess if it isn't auto-committing, we should at least add a step that checks that the code _is_ in sync w/ black, and add a note to README/Contributing on how to do this. I just want to make sure I understand our strategy going forward, as I'm excited about this change!

Just pushed a commit that contains documentation on how to use black. As for adding a step that checks that black is in sync, that's already included [in the pipeline](https://github.com/mintapi/mintapi/pull/333/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR34). Before I just meant that the pipeline won't fix it for you, it just will check to make sure that no fixing is needed.

mrooney:
> > @jaron-l makes sense, I guess if it isn't auto-committing, we should at least add a step that checks that the code _is_ in sync w/ black, and add a note to README/Contributing on how to do this. I just want to make sure I understand our strategy going forward, as I'm excited about this change!
> 
> Just pushed a commit that contains documentation on how to use black. As for adding a step that checks that black is in sync, that's already included [in the pipeline](https://github.com/mintapi/mintapi/pull/333/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR34). Before I just meant that the pipeline won't fix it for you, it just will check to make sure that no fixing is needed.

That makes sense! Sorry, I thought you meant the step fails to check :) LGTM, feel free to update and merge! Thanks :)

mrooney (dismissed):

jaron-l:
@mrooney looks like your approval expires if I update the branch. Can you approve/merge this before another merge?

mrooney (approved):

mrooney:
Done, thanks for your hard work on keeping this up to date :)

Issue #334, opened 2021-11-07
Mintapi fails when installing on python2.7 due to dependencies requiring python>=3.6


matt199x:
When running "pip install mintapi" the following error occurs:

Collecting oathtool (from mintapi)
Could not find a version that satisfies the requirement oathtool (from mintapi) (from versions: )
No matching distribution found for oathtool (from mintapi)

jaron-l:
This is likely a internet connection issue or something like that from pip
not being able to query PyPI. You'll probably be able to find the cause of
the issue from somewhere else in the error stack trace.

On Sun, Nov 7, 2021, 11:48 AM matt199x 'notifications at github.com'
<jaron-github+reply+ADGNKUFL4BVDG5R7K36JDSV7SPZH3EVBNHHD4ZIILA=

matt199x:
I have been able to install several other items using "pip install" today. I tried on another raspberry bi and it has the same issue.

mrooney:
Is there a full stack trace you can paste as per the bug report template?

matt199x:
running "sudo pip3 install mintapi" instead of "pip install mintapi" fixed the issue.

jaron-l:
I don't think this is a mintapi issue but rather something about your
python setup. I was just able to install mintapi on a raspberry pi without
sudo or any other hiccups. The full stack trace should lead you to what the
issue is.

On Sun, Nov 7, 2021, 7:03 PM matt199x 'notifications at github.com'
<jaron-github+reply+ADGNKUEX73MRKLVVPCJCFGN7SRMGLEVBNHHD4ZIILA=

matt199x:
`pi@raspberrypi:~ $ pip install mintapi
Looking in indexes: https://pypi.org/simple, https://www.piwheels.org/simple
Collecting mintapi
  Downloading https://files.pythonhosted.org/packages/f6/94/6542c1a3d82ad599aa2bfcea4cfa10c84e99bd0ace2a830888a9993d8c84/mintapi-1.54.tar.gz
Collecting future (from mintapi)
  Downloading https://files.pythonhosted.org/packages/45/0b/38b06fd9b92dc2b68d58b75f900e97884c45bedd2ff83203d933cf5851c9/future-0.18.2.tar.gz (829kB)
    100% |████████████████████████████████| 829kB 324kB/s 
Requirement already satisfied: mock in /usr/lib/python2.7/dist-packages (from mintapi) (2.0.0)
Collecting oathtool (from mintapi)
[...]

PR #335, opened 2021-11-09
fix: :bug: Set minimum python version to 3.6

jaron-l:
* remove python2 code
* remove pandas import checking as it is a dependency

fixes: #334

I also note, that we could potentially find the oathtool version that supports python2.7, but I feel obliged to recommend against that seeing as python2.7 is EOL.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -19,11 +19,6 @@ import email.header
 import sys  # DEBUG
 import warnings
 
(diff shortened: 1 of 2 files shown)


burkematthew (approved):

mrooney (approved):
Very reasonable, thanks for the cleanup!

PR #336, opened 2021-11-10
feat: :sparkles: Expose sign_in as a public method to allow custom WebDrivers

jaron-l:
* show deprecation warning on get_web_driver
* refactor login_and_get_token to cover get_web_driver
* document how to use sign_in with custom WebDriver

refs: #330

This is a proposal to replace #330 which will allow the user to specify custom WebDrivers so we don't have to explicitly expose every requested feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,23 @@ If you're running mintapi in a server environment on an automatic schedule, cons
 
 If you need to download the chromedriver manually, be sure to get the version that matches your chrome version and make the chromedriver available to your python interpreter either by putting the chromedriver in your python working directory or inside your `PATH` as described in the [python selenium documentation](https://www.selenium.dev/selenium/docs/api/py/index.html#drivers).
 
+### General Automation Scenarios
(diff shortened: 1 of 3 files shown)


mrooney:
Looks pretty cool to me, I can't really think of any downside. It doesn't seem like it adds any real complexity to the code (you did a bunch of cleanup, actually :), and makes more setups possible. Nice!

mrooney:
(That said, I do think if this replaces #330, we should still improve the README for simpler cases where the user just wants to know how to specify PATH properly in a cron or script)

jaron-l:
> (That said, I do think if this replaces #330, we should still improve the README for simpler cases where the user just wants to know how to specify PATH properly in a cron or script)

I just added a commit that should address this. Let me know what you think.

jaron-l:
@mrooney just in case you didn't see, I addressed your previous comment. Let me know if you have any more feedback.

mrooney (approved):
Awesome, thanks for this slick improvement, LGTM!

PR #338, opened 2021-11-11
remove python 3.6 support

mrooney:
Just making this PR in advance since python 3.6 is EOL on Dec 23rd. I think it would also be fine to merge early since it seems only helpful to encourage upgrades, and I don't foresee any py3.6 issues coming up in the next few weeks, and it probably wouldn't make sense to fix anyway :) I'm also fine with waiting if that seems better.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: mintapi build 
+name: mintapi build
 
 on: [push, pull_request]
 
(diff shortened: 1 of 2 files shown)


jaron-l:
I'm kinda of the mind of not enforcing upgrade until it becomes a developer burden, .i.e we have to support alternate code or dependencies require it. Since we have 3.6 in the pipeline, we should be notified this time if either case happens as either the tests will fail or the pipeline won't be able to install a dependency.

mrooney:
> I'm kinda of the mind of not enforcing upgrade until it becomes a developer burden, .i.e we have to support alternate code or dependencies require it. Since we have 3.6 in the pipeline, we should be notified this time if either case happens as either the tests will fail or the pipeline won't be able to install a dependency.

That's a fair point, I'll close it for now :)

PR #340, opened 2021-11-12
Add command-line option to exclude credit inquiries from credit report data

### LINKED ISSUE #326: Add command-line option to control if inquiries are excluded
If a credit report is requested, there are 3 different types of details included automatically: credit inquiries, credit accounts, and credit utilization history.  It'd be nice to have a way to isolate a certain type of that data.  To preserve existing functionality, we should add an "exclude" option for credit inquiries.

burkematthew:
The changes made to `api.py` ended up being more involved only to provide a better foundation for the tests needed.  If there are any suggestions to better accommodate the test scenario, I'm all ears, but the test is working correctly and the results of the command-line option are as expected as well.

This is to resolve #326 .

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -1334,19 +1334,30 @@ class Mint(object):
         return self.driver.get(MINT_CREDIT_URL)
 
     def _get_credit_reports(self, limit, credit_header):
-        return self.get("{}/v1/creditreports?limit={}".format(MINT_CREDIT_URL, limit), headers=credit_header).json()
+        return self.get(
+            "{}/v1/creditreports?limit={}".format(MINT_CREDIT_URL, limit),
+            headers=credit_header,
+        ).json()
 
     def _get_credit_details(self, url, credit_header):
         return self.get(url.format(MINT_CREDIT_URL), headers=credit_header).json()
 
     def get_credit_inquiries(self, credit_header):
-        return self._get_credit_details("{}/v1/creditreports/0/inquiries", credit_header)
+        return self._get_credit_details(
+            "{}/v1/creditreports/0/inquiries", credit_header
+        )
 
     def get_credit_accounts(self, credit_header):
-        return self._get_credit_details("{}/v1/creditreports/0/tradelines", credit_header)
+        return self._get_credit_details(
+            "{}/v1/creditreports/0/tradelines", credit_header
+        )
 
     def get_credit_utilization(self, credit_header):
-        return self.__process_utilization(self._get_credit_details("{}/v1/creditreports/creditutilizationhistory", credit_header))
+        return self.__process_utilization(
+            self._get_credit_details(
+                "{}/v1/creditreports/creditutilizationhistory", credit_header
+            )
+        )
 
     def __process_utilization(self, data):
         # Function to clean up the credit utilization history data
@@ -1810,8 +1821,9 @@ def main():
     elif options.credit_score:
         data = mint.get_credit_score()
     elif options.credit_report:
-        data = mint.get_credit_report(details=True,
-                                      exclude_inquiries=options.exclude_inquiries)
+        data = mint.get_credit_report(
+            details=True, exclude_inquiries=options.exclude_inquiries
+        )
 
     # output the data
     if options.transactions or options.extended_transactions:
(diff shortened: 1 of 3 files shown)


tests/test_driver.py
```diff
@@ -181,6 +181,29 @@ def test_detailed_transactions_with_parents(self, mock_get_categories):
...
+
+    @patch.object(mintapi.Mint, '_get_api_key_header')
+    @patch.object(mintapi.Mint, '_load_mint_credit_url')
+    @patch.object(mintapi.api, 'get_web_driver')
+    @patch.object(mintapi.Mint, '_get_credit_reports')
+    @patch.object(mintapi.Mint, 'get_credit_inquiries')
+    @patch.object(mintapi.Mint, 'get_credit_accounts')
+    @patch.object(mintapi.Mint, 'get_credit_utilization')
+
+    def test_exclude_inquiries(self, mock_get_api_key_header, mock_load_mint_credit_url, mock_driver, mock_get_credit_reports, mock_get_credit_inquiries, mock_get_credit_accounts, mock_get_credit_utilization):
```
jaron-l, line 193: I'm not saying it's wrong but... what is the intended scope of this test? Seems like you could easily just mock `get_credit_accounts` and `get_credit_utilization` and the assert would still work. Seems like a lot of extra set up.

mintapi/api.py
```diff
@@ -1117,34 +1117,44 @@ def get_credit_report(self, limit=2, details=True):
...
+        return self.get(url.format(MINT_CREDIT_URL), headers=credit_header).json()
+
+    def get_credit_inquiries(self, credit_header):
+        return self._get_credit_details('{}/v1/creditreports/0/inquiries', credit_header)
+
+    def get_credit_accounts(self, credit_header):
+        return self._get_credit_details('{}/v1/creditreports/0/tradelines', credit_header)
+
+    def get_credit_utilization(self, credit_header):
+        return self.process_utilization(self._get_credit_details('{}/v1/creditreports/creditutilizationhistory', credit_header))
```
jaron-l, line 1157: Do you think it would be harmful to go ahead and mark `process_utilization` as a private function?

jaron-l (changes requested):

tests/test_driver.py
burkematthew: @jaron-l the scope of this test is to verify that when passing `exclude-inquiries` that the resulting object does not return.  I'll take a look again and see if I can simplify the test, but because there are so many requests associated with getting credit detail, it needed several mocks.  But let me see what I can do.

mintapi/api.py
burkematthew: I'll take a look and see.  My inclination is that it would be a good idea to mark that `process_utilization` as private.  The purpose is to serve as a simplification of logic surrounding getting credit utilization data.  It really has no external purpose.  I'll verify throughout the source and so long as nothing else stands out, I'll implement.

tests/test_driver.py
burkematthew: @jaron-l I looked at this again.  Looks like I can remove the mock of `get_credit_inquiries` since we're excluding that data anyways, so thanks for that assist.  All of the others are necessary, though, since we are accessing another URL and so there is more that goes into the `get_credit_report` method compared to others.

burkematthew:
@jaron-l ready for your review again.

burkematthew:
@jaron-l I've got merge conflicts resolved and this is now up to the standards of the new linter.

tests/test_driver.py
jaron-l: Could you just mock the `Mint` class and assert that `get_credit_inquires` was not called? Would simplify the code a lot. Also, I noticed that your patching is backwards. Python decorators work closest to function act first, i.e. in your case, your patch for `get_credit_utilization` will show up in the first argument where you have `mock_get_api_key_header`.

jaron-l: Ah nevermind, that won't work. Of you could just fix the mocks and if I can't think of a clearer way we'll leave it.

burkematthew: @jaron-l fixed the order of the mocks.

jaron-l: I noticed that you don't actually modify any of the mock objects so you can probably delete all the `return_value` modifications. I'll go ahead and do that, push, and merge.

jaron-l (approved):

Issue #227, opened 2020-10-06
invest_json() changes


ghost:
looks like mint has deprecated the old https://mint.intuit.com/investment.event api and moved to 
https://mint.intuit.com/investment-new.event. Is this repo still active? Lots of cool new things here. I'm considering becoming a contributor.

mrooney:
Yes, it is quite active! We always welcome contributions and PRs are
regularly merged :) Are you comfortable submitting a PR using the new API?

ljc-codes:
That sounds awesome! Yeah, little busy with life stuff right now but I'll give it a shot in the coming weeks and shoot you guys a pull request! 

Best
Tiger Shark

burkematthew:
It looks like we do not currently support `get_invests_json` from the command-line, which is the method that actually makes this request.  @mrooney are you good with this change including adding support for `get_invests_json` from the command-line, while also switching over to the new endpoint?

mrooney:
> It looks like we do not currently support `get_invests_json` from the command-line, which is the method that actually makes this request. @mrooney are you good with this change including adding support for `get_invests_json` from the command-line, while also switching over to the new endpoint?

Sounds like a great approach all around, thanks!

burkematthew:
@jaron-l @mrooney I've got this working well on the draft PR submitted.  However, I do want to point out that the data structure does change, which could be a breaking change for some users.  How do we want to approach this?  I could simply add a new method that handles the new endpoint, while preserving the old method for those that use it.  In addition, since command-line doesn't exist at the moment at all for investment data, I could point the command-line option to the new endpoint.

Thoughts?

jaron-l:
> @jaron-l @mrooney I've got this working well on the draft PR submitted. However, I do want to point out that the data structure does change, which could be a breaking change for some users. How do we want to approach this? I could simply add a new method that handles the new endpoint, while preserving the old method for those that use it. In addition, since command-line doesn't exist at the moment at all for investment data, I could point the command-line option to the new endpoint.
> 
> Thoughts?

[...]

PR #342, opened 2021-11-12
✨ Adding support for the new investments JSON and adding command-line support

burkematthew:
Addressing #227.


burkematthew:
@mrooney let me know your thoughts on this one!

burkematthew:
Note: need to explore further linter issues.

mintapi/api.py
```diff
@@ -846,6 +851,18 @@ def get_invests_json(self):
         else:
             logger.error("FAIL2")
 
+    def get_investment_data_new(self):
```
jaron-l, line 854: Can we leave off the suffix "new"? I generally dislike time linked function names.

mintapi/api.py
```diff
@@ -846,6 +851,18 @@ def get_invests_json(self):
...
+        investments = self.call_investments_endpoint()["Investment"]
+        for i in investments:
+            i.pop("metaData", None)
+        return investments
+
+    def call_investments_endpoint(self):
```
jaron-l, line 860: Can we make this a private function? I think a goal for mintapi would be to be clear about what the public api is so we know what is a breaking change or not.

mintapi/api.py
```diff
@@ -846,6 +851,18 @@ def get_invests_json(self):
         else:
             logger.error("FAIL2")
 
+    def get_investment_data_new(self):
+        investments = self.call_investments_endpoint()["Investment"]
+        for i in investments:
+            i.pop("metaData", None)
```
jaron-l, line 857: Why does the metaData need to be removed?

jaron-l (changes requested):

mintapi/api.py
burkematthew: I reviewed the metaData object again and decided that the "lastUpdatedDate" field would actually be useful to flatten.  Everything else isn't useful information and by including it, we lose the ability to send investment data to a CSV.

burkematthew:
@jaron-l back to you for review.

tests/test_endpoints_sign_in.py
```diff
@@ -0,0 +1,56 @@
...
+
+def test_sign_in(get_mint_driver: mintapi.Mint):
+    mintapi.api._sign_in(
+        USERNAME, PASSWORD, get_mint_driver.driver,
+        MFA_METHOD, MFA_TOKEN, INTUIT_ACCOUNT,
+    )
+    assert get_mint_driver.driver.current_url.startswith("https://mint.intuit.com/overview.event")
+   
+
+def test_investment_endpoint(do_sign_in: mintapi.Mint):
```
jaron-l, line 52: I rearranged the sign_in tests to make them a little bit easier to work with. I then added a test for your endpoint. This test fails despite it replicating what your added test on this branch. Do you know why?

jaron-l (changes requested):

tests/test_endpoints_sign_in.py
burkematthew: @jaron-l it looks like you got it to work now.  However, I will note that it looks like you're using an incorrect formatting for declaring a Python method private.  It's my understanding that it requires two underscores, not one.  Just curious if you intended to remove the classification as private?

jaron-l: Yeah, I guess I didn't need it anymore. I started by writing tests that call the endpoint functions themselves and so I needed them to be only semi-private so they were testable. They can probably be reverted. I'll fix it. Generally though, I think we should have some sort of definition of what should be public, semi-private, and private as that is a debate in software engineering.

But I don't think you see the issue that I was trying to raise. When I call `__call_investment_endpoint`, it doesn't have he investment key that is required in `get_investment_data`. Now that I think about it, it's probably because I don't have my investments listed in mintapi. We probably need to fail safely from that.

jaron-l: Actually, I might want to argue that both `_flatten_utilization` and `_process_utilization` should be semi-private as ideally they would be testable code. I agree that `__call_investment_endpoint` probably can be private.

burkematthew: I'm fine on the privacy level mentioned related to utilization. No big deal there. 

And let me know if you still need me to look at the issue you're having.

jaron-l: Can you add in the error handling for if the "Investment" key is not found in the return json?

jaron-l: Our command line code doesn't actually handle the case that data is None, so you'll need to add that too. #problems-with-duck-typing

jaron-l (changes requested):

burkematthew:
@jaron-l ready for you!

jaron-l (approved):

Issue #347, opened 2021-11-20
mintapi 1.57 hangs on overview screen with --extended-transactions


apiology:
**Describe the bug**
mintapi 1.57 hangs on overview screen

What version of MintAPI are you using? 1.57
What command[s] did you run / steps to reproduce?* mintapi --use-chromedriver-on-path --extended-transactions my-username my-password
Stacktrace/error received:

After waiting five minutes, I hit control-c  and get this:

```
C-c C-cTraceback (most recent call last):
  File "/Users/broz/.pyenv/versions/mylibs/bin/mintapi", line 8, in <module>
    sys.exit(main())
  File "/Users/broz/.pyenv/versions/3.8.5/envs/mylibs/lib/python3.8/site-packages/mintapi/api.py", line 1380, in main
    data = mint.get_detailed_transactions(
  File "/Users/broz/.pyenv/versions/3.8.5/envs/mylibs/lib/python3.8/site-packages/mintapi/api.py", line 835, in get_detailed_transactions
[O    df = pd.DataFrame(self.add_parent_category_to_result(result))
  File "/Users/broz/.pyenv/versions/3.8.5/envs/mylibs/lib/python3.8/site-packages/mintapi/api.py", line 850, in add_parent_category_to_result
    parent = self.get_category_object_from_id(transaction['categoryId'])['parent']
  File "/Users/broz/.pyenv/versions/3.8.5/envs/mylibs/lib/python3.8/site-packages/mintapi/api.py", line 1084, in get_category_object_from_id
    categories = self.get_categories()
  File "/Users/broz/.pyenv/versions/3.8.5/envs/mylibs/lib/python3.8/site-packages/mintapi/api.py", line 985, in get_categories
    response = self.make_post_request(url=self.build_bundledServiceController_url(),
  File "/Users/broz/.pyenv/versions/3.8.5/envs/mylibs/lib/python3.8/site-packages/mintapi/api.py", line 617, in make_post_request
    response = self.post(url=url, data=data, headers=JSON_HEADER)
  File "/Users/broz/.pyenv/versions/3.8.5/envs/mylibs/lib/python3.8/site-packages/mintapi/api.py", line 614, in post
    return self.driver.request('POST', url, **kwargs)
  File "/Users/broz/.pyenv/versions/3.8.5/envs/mylibs/lib/python3.8/site-packages/seleniumrequests/request.py", line 232, in request
    self.add_cookie(cookie_dict)
[...]

PR #348, opened 2021-11-20
🐛 Fix performance degradation on extended transactions

burkematthew:
This is to address the performance reported in #347 

Previously, ever iteration of `get_category_object_from_id` would make a `GET` request for the categories.  In the case of pulling extended transaction data, that's one request per transaction, which can add up to a LOT of unnecessary requests.  The idea behind this change is to pull the category data in `get_budgets` and `add_parent_category_to_result` and then pass it to `get_category_object_from_id`.


burkematthew:
@mrooney @jaron-l can we get this one moving? Pulling extended data is pretty degraded without this change.

mrooney (approved):
LGTM, thanks!

Issue #327, opened 2021-10-28
Add command-line option to control if credit accounts are excluded


burkematthew:
If a credit report is requested, there are 3 different types of details included automatically: credit inquiries, credit accounts, and credit utilization history.  It'd be nice to have a way to isolate a certain type of that data.  To preserve existing functionality, we should add an "exclude" option for credit accounts.

jaron-l:
> Same question, does it currently require multiple requests to get this info? I ask because if not, I'd lean towards not complicating mintapi code to do processing of data that can be easily extracted on the client side.

It does, see in [get_credit_report](https://github.com/mintapi/mintapi/blob/56c01912309dad557bd70125645ef8af994c0382/mintapi/api.py#L1151).

PR #351, opened 2021-11-26
✨ Feature: New command-line option to exclude credit accounts from credit report

burkematthew:
Addresses #327


mintapi/api.py
```diff
@@ -1301,7 +1305,7 @@ def get_credit_report(self, limit=2, details=True, exclude_inquiries=False):
 
         self._load_mint_credit_url()
 
-        credit_report["reports"] = self._get_credit_reports(limit, credit_header)
+        # credit_report["reports"] = self._get_credit_reports(limit, credit_header)
```
jaron-l, line 1308: Is this removal intentional?

jaron-l (changes requested):

mintapi/api.py
burkematthew: Nope. Got caught in my last test. Thanks for that catch.

jaron-l (approved):

PR #352, opened 2021-11-26
feat(api): :sparkles: Make mfa_input_callback available for soft-token method

jaron-l:
* move getpass, oathtool, and keyring hidden imports to top of file

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,14 +19,13 @@ setuptools.setup(
     url="https://github.com/mintapi/mintapi",
     install_requires=[
         "configargparse",
-        "future",
-        "mock",
         "oathtool",
         "pandas>=1.0",
         "requests",
         "selenium",
         "selenium-requests>=1.3.3",
         "xmltodict",
+        "keyring",
     ],
     python_requires=">=3.6",
     entry_points=dict(
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ make calls to retrieve account/budget information.  We recommend using the
                        # if mintapi detects an MFA request, it will trigger the requested method
                        # and prompt on the command line.
     mfa_input_callback=None,  # see MFA Methods section
-                              # used with mfa_method = 'sms' or 'email'
+                              # can be used with any mfa_method
                               # A callback accepting a single argument (the prompt)
                               # which returns the user-inputted 2FA code. By default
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -19,14 +19,13 @@
     url="https://github.com/mintapi/mintapi",
     install_requires=[
         "configargparse",
-        "future",
-        "mock",
```
mrooney, line 23: These ones aren't actually used?

mrooney:
Makes sense overall. I think the original idea was that these options might not be mainstream (they obviously weren't when they were added) and that it only made sense to require them if people were going to use them. But maybe these are all popular options now or it is easier to require them all?

Also, now that we have more imports, does it make sense to alphabetize them? That can certainly be in a follow up PR.

setup.py
jaron-l: They're included in the standard library of python 3.

jaron-l: Or rather, mock is. Future is no longer used as it was for python 2 to 3 compatibility.

jaron-l:
> Makes sense overall. I think the original idea was that these options might not be mainstream (they obviously weren't when they were added) and that it only made sense to require them if people were going to use them. But maybe these are all popular options now or it is easier to require them all?
> 
> Also, now that we have more imports, does it make sense to alphabetize them? That can certainly be in a follow up PR.

I think it is much easier to just require them on install. Pip will just handle it then. There's really no disadvantage at all to require it on install versus having to manually install packages as you find certain features require them. Plus, the latter has no dependency management system so any advantage in not requiring would quickly get more complicated at scale. The only real optional dependency management system I've seen is in python-poetry.

I don't really see an advantage to alphabetizing them. I've been generally organizing imports by built-ins, external packages, and internal modules as that is somewhat a common practice so you know when to careful about circular imports.

mrooney:
> I don't really see an advantage to alphabetizing them. I've been generally organizing imports by built-ins, external packages, and internal modules as that is somewhat a common practice so you know when to careful about circular imports.

That makes sense! My main desire is that if someone is looking to add or remove an import, there's a right place and not an "I'll throw it here, I guess"-place. It's a lot nicer to visually scan too. I can do that in a follow up PR though.

mrooney (approved):

PR #354, opened 2021-11-30
release version 1.58

mrooney:
Party time, lots of goodies here! I believe these are all appropriate for a 1.x version bump.

1.58
---
- expose sign_in as a public method to allow custom WebDrivers (#336)
- add command-line option to exclude credit inquiries from credit report data (#340)
- add command-line option to exclude credit accounts from credit report (#351)
- fix performance degradation on extended transactions (#348)
- remove support for Python 2.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+1.58
+---
+- expose sign_in as a public method to allow custom WebDrivers (#336)
+- add command-line option to exclude credit inquiries from credit report data (#340)
+- add command-line option to exclude credit accounts from credit report (#351)
+- fix performance degradation on extended transactions (#348)
+- remove support for Python 2.x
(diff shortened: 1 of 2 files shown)


burkematthew:
@mrooney might be good to wait for #356 so we knock out all 3 of those related options in the same release.

mrooney:
> @mrooney might be good to wait for #356 so we knock out all 3 of those related options in the same release.

There will always be something to wait for, I'm a fan of releasing earlier and more often to get feedback earlier, vs trying to figure out which change caused a bug. I didn't actually mean to wait this long and bundle this many things in one version, haha.

mrooney:
@burkematthew that said I appreciate the suggestion and would probably take it if there were only a thing or two in the 1.58 release :)

Issue #328, opened 2021-10-28
Add command-line option to control if utilization history is excluded


burkematthew:
If a credit report is requested, there are 3 different types of details included automatically: credit inquiries, credit accounts, and credit utilization history. It'd be nice to have a way to isolate a certain type of that data. To preserve existing functionality, we should add an "exclude" option for credit utilization history.

PR #356, opened 2021-11-30
✨ Adding new option to exclude credit utilization history data

burkematthew:
Addresses #328

### DIFF
--- a/tests/test_driver.py
+++ b/tests/test_driver.py
@@ -226,6 +226,21 @@ class MintApiTests(unittest.TestCase):
         )
         self.assertFalse("accounts" in credit_report)
 
+    @patch.multiple(
+        mintapi.Mint,
+        _get_api_key_header=DEFAULT,
+        _load_mint_credit_url=DEFAULT,
+        _get_credit_reports=DEFAULT,
+        get_credit_inquiries=DEFAULT,
+        get_credit_accounts=DEFAULT,
+    )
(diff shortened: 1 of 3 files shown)


tests/test_driver.py
```diff
@@ -226,6 +226,21 @@ def test_exclude_accounts(self, **_):
...
+        _get_credit_reports=DEFAULT,
+        get_credit_inquiries=DEFAULT,
+        get_credit_accounts=DEFAULT,
+    )
+    def test_exclude_utilization(self, **_):
+        mint = mintapi.Mint()
+        credit_report = mint.get_credit_report(
+            limit=2, details=True, exclude_utilization=True
+        )
+        self.assertFalse("utilization" in credit_report)
```
mrooney, line 242: Nice work! Maybe this is how the other tests work and I missed it, but if the key changed from "utilization", this test would still keep on passing under all scenarios. I'd prefer that we added a line to each exclusion test that verifies the key *does* exist without the exclusion option specified, and then that it goes away when you do exclude it. What do you think?

mrooney (changes requested):

tests/test_driver.py
burkematthew: @mrooney good point.  I like it!  I'll do that for this test and the others as well.

burkematthew:
@mrooney back over to you!

mrooney (approved):

PR #358, opened 2021-12-03
refactor: :recycle: Split up cli code from api.py

jaron-l:
api.py is almost 2000 lines of code. I think it might help keep the repo clean and easier to understand if we start breaking out the cli and sign-in code away from the rest of the interface. This pull request shows how we could break out the cli code from api.py.


mrooney:
It is definitely big, I'm in favor of this direction!

burkematthew:
This looks awesome and makes sense! Nice job!

jaron-l:
@mrooney, so I was going through the README.md to make sure I didn't break any entry points that people might be using and I came across the last line in the README.md that says this:
```markdown
If you need to avoid using pip or setup.py, you can also clone/download
this repository and run: python mintapi/api.py
```

As far as I know (please let me know if you know of a way), there isn't a way to support uninstalled python package imports without manually adding paths to `sys.path`, which doesn't scale/adapt well to package changes.

In addition, I then realized that the line in the README.md wouldn't work anyways. It doesn't install the package dependencies. The person would first need to do `pip install .` inside the cloned repository to get the dependencies, which goes against the first part of the sentence anyways as it uses both pip and setup.py. We could document the way to add the repository to path, but as that doesn't scale, it's likely to break. In addition, I feel that I would be outside the scope of a normal use case. The person that needs to use mintapi in such a specific way would likely already know how to do that. I don't think we should recommend it to the average user.

I want to recommend that we just remove the line and only support installed versions of mintapi. If people need to use it in a different way, they can figure it out as it is the same problem for any python package and adds burden to us.

I want to suggest that we just remove that line and not directly support uninstalled versions of mintapi.

mrooney:
@jaron-l I'm great with that! Probably it was put there originally when it first made it on to pypi. I agree with all your points :)

README.md
```diff
@@ -26,7 +26,7 @@ pip install mintapi
 
 From the command line, the most automated invocation will be:
 
-    python mintapi/api.py --keyring --headless you@example.com
+    mintapi --keyring --headless you@example.com
```
mrooney, line 29: Does this still need the python binary, or am I missing how this would work with just a plain 'mintapi'? Overall looks awesome, this was the only part I was unclear on. Thanks for this undertaking!

jaron-l: It works both ways with the caveat that if you put the python binary there, you need to have a local repo of mintapi in your current working directory as python will treat it like a path. I think that will be an abnormal situation as most will install mintapi from PyPI. 

If you don't put the python binary there, your terminal process will search on your `PATH` to find mintapi and it will be there if you have mintapi installed globally or in an activated virtual environment. This is the way that we already recommend as you can see from the CLI section at the end of the README.md.

mrooney: Ah cool, I had forgotten that, thanks! Makes sense.

mrooney (approved):
Approving and if it needs a re-approval after resolving the conflict, just let me know.

jaron-l:
@mrooney just merged main into this branch. Should be good to go.

mrooney (approved):

mrooney:
Awesome, thanks for this great refactor!

PR #359, opened 2021-12-04
🧹 Convert to the updated categories endpoint

burkematthew:
* Converted to the new categories endpoint
* Added support for pulling categories from the command line
* Reduced redundancies in formatting budget category data


burkematthew:
@mrooney thoughts on this one?

mintapi/api.py
```diff
@@ -865,6 +865,19 @@ def __call_investments_endpoint(self):
             headers=self._get_api_key_header(),
         ).json()
 
+    def get_categories(self):
+        try:
+            categories = self.__call_categories_endpoint()["Category"]
+        except Exception:
```
jaron-l, line 871: We should avoid hiding exceptions like this. This is not actually handling the error. It is hiding that it happened and it will raise a more ambiguous error later in add_parent_category_to_result. Either let the error happen or, if there is a way to proceed and give a meaningful result, do that.

mintapi/api.py
```diff
@@ -1575,6 +1557,14 @@ def parse_arguments(args):
             ("--imap-test",),
             {"action": "store_true", "help": "Test imap login and retrieval."},
         ),
+        (
+            ("--categories",),
```
jaron-l, line 1561: Might be good to say that this is talking about budget categories versus all categories in transactions.

mintapi/api.py
```diff
@@ -1272,32 +1243,43 @@ def mos_to_yrmo(mos_frm_zero):
...
         return budgets
 
+    def __format_budget_categories(self, budget, categories):
+        category = self.get_category_object_from_id(budget["cat"], categories)
+        budget["cat"] = category["name"]
+        parent = self._find_parent_from_category(category)
+        budget["parent"] = parent["name"]
+        return budget
+
     def get_category_object_from_id(self, cid, categories):
```
jaron-l, line 1257: I think we need a test for this function to make sure we didn't break it.

tests/test_driver.py
```diff
@@ -24,36 +24,34 @@
     }
 ]
 
-category_example = {
-    708: {
+category_example = [
```
jaron-l, line 27: Does this change mean this PR is a breaking change?

jaron-l (changes requested):

tests/test_driver.py
burkematthew: No.  All we pulled the categories for previously was to get the budget category and the parent category for transactions.  The code change involved in this PR accounts for the fact that the data returned in the new endpoint is in a different structure than the original endpoint.

mintapi/api.py
burkematthew: Well it's a way to export all the categories associated with the user's Mint account.  It's not just budget categories.  That being said, I can clarify that this is a standalone option to pull your categories from Mint.

mintapi/api.py
burkematthew: Roger that.  I'll reconsider this section.

burkematthew:
@jaron-l test added and your other comments addressed.  Let me know if you have anything else for me to revise.

mrooney (approved):
LGTM! Now that the test fixture data is getting larger, we should move that json data/etc to files inside the test directory, so the test file itself is more readable, but I am happy to do that in a follow-up PR.

Any remaining concerns @jaron-l ?

burkematthew:
@jaron-l anything else on this one?

jaron-l (approved):

PR #360, opened 2021-12-04
✨ Use keyring for IMAP password

burkematthew:
To enhance the security potential and increase consistency of how we handle password, I have created the following:
* A common method to handle password logic based on a given username.
* Apply the new method to Mint Username and Password
* Apply the new method to IMAP Username and Password in `mfa_method="email"` and if `imap_account`.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -80,7 +80,11 @@ def get_email_code(
             DeprecationWarning,
         )
     code = None
-    imap_client = imaplib.IMAP4_SSL(imap_server)
+    try:
+        imap_client = imaplib.IMAP4_SSL(imap_server)
+    except imaplib.IMAP4.error:
+        logger.error("ERROR: Unable to establish IMAP Client")
+        return ""
 
     try:
         rv, data = imap_client.login(imap_account, imap_password)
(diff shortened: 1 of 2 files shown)


mintapi/api.py
```diff
@@ -1744,29 +1748,48 @@ def initiate_account_refresh(email, password):
     return mint.initiate_account_refresh()
 
 
+def handle_password(type, prompt, email, password, use_keyring=False):
+    if not password:
+        # If we don't yet have a password, try prompting for it
+        password = keyring.get_password(type, email)
```
mrooney, line 1754: Why do we try to get the password from the keyring when use_keyring is False?

mrooney:
Seems like a great idea!

mintapi/api.py
burkematthew: That's a great question. Previously, when we supported python V2, we checked if keyring was installed and if a password was supplied or not. Now, keyring is always imported, so we are really just seeing if a password was supplied or not.  However, now looking at the code again, we probably should have changed the conditional when we removed support for v2 as python was always installed. At this point, it'd make sense to only use keyring if the user wants to. In addition, I think it might make sense in v2 to ALWAYS use keyring for password management unless the user submits the password at runtime. 

Thoughts?

mrooney: Makes sense. Just checking for `use_keyring` makes sense for this PR, and I also agree changing to some more modern defaults for v2 sounds good, including keyring, and we can add a --no-keyring option if someone really wants a prompt.

mintapi/api.py
```diff
@@ -1744,29 +1748,48 @@ def initiate_account_refresh(email, password):
...
+    password = handle_password(
+        "mintapi", "Mint password: ", email, password, options.keyring
+    )
 
-    if options.keyring:
-        # If keyring option is specified, save the password in the keyring
-        keyring.set_password("mintapi", email, password)
+    if mfa_method == "email" and imap_account:
+        imap_password = handle_password(
+            "imap",
```
mrooney, line 1787: Do you think it might be worth namespacing this to "mintapi_imap" or similar? If this just ends up in the keyring as "imap" that's a bit generic and could potentially collide with something else. Overall looks great!

mrooney (changes requested):

mintapi/api.py
burkematthew: @mrooney that's a good point.  I'll change that namespace.

jaron-l (changes requested):

mintapi/api.py
```diff
@@ -80,7 +80,11 @@ def get_email_code(
             DeprecationWarning,
         )
     code = None
-    imap_client = imaplib.IMAP4_SSL(imap_server)
+    try:
+        imap_client = imaplib.IMAP4_SSL(imap_server)
+    except imaplib.IMAP4.error:
+        logger.error("ERROR: Unable to establish IMAP Client")
+        return ""
```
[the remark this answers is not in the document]
burkematthew: @jaron-l I do agree, but I think that should be done as a follow-up instead of as a part of this particular PR.  I simply added this here as it was an opportunity for unexpected behavior as it stood with using keyring for the IMAP password.

I can do the follow-up work on raising true exceptions for these IMAP errors.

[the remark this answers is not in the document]
jaron-l: Sounds good to me. I went ahead and made #362 to track this.

jaron-l (approved):

mrooney (approved):

PR #361, opened 2021-12-06
Move Black Linter to separate GitHub step

burkematthew:
For contributors, it is helpful to evaluate the GitHub checks in a more modular manner.  In particular, it is helpful to have a separate step for running the Black linter from the actual tests associated with each version of Python.


jaron-l (approved):

mrooney:
Very cool, thanks!

PR #366, opened 2021-12-09
refactor: :recycle: Split sign in

jaron-l:
* move sign in code to sign_in.py

Let me know what you think!


jaron-l:
@mrooney, Just fyi, I intentionally created two branches that I merged together for this PR when doing the split of api.py (you'll be able to see this if you look at the graph of the commits on this branch). I did this to maintain line history for the new sign_in_mod.py file. If you want to maintain that line history, you'll need to merge this with a merge commit instead of the normal squash commit method. If you don't care for that or think that it's fine that git blame makes it look like I wrote sign_in_mod.py, then you can squash as normal. Up to you.

mrooney (changes requested):
Looks awesome, thanks for taking this on! I think it will be great to have sign in code isolated :) And, I don't think it is a problem if the line attribution changes.

Two minor thoughts:
* can we just call the file `sign_in`? sign_in_mod feels a little awkward, and we don't have `cli_mod` or `cli_file` or anything like that.
* if we are going to import `_create_web_driver_at_mint_com`, it isn't really internal and we may as well just remove the beginning underscore.

What do you think?

jaron-l:
> * can we just call the file `sign_in`? sign_in_mod feels a little awkward, and we don't have `cli_mod` or `cli_file` or anything like that.

`sign_in_mod.py` and the `sign_in` function inside `sign_in_mod.py` cannot have the same names with the current import configuration. Because we import the sign_in function in `__init__.py` there is a name conflict with which module you are referencing if you want to do `from mintapi import sign_in` or any variation of that import statement. I would think that the function should take precedence in keeping the name since it is part of our public API and changing it would result in a breaking change. I'm down with if there is a better name for the file.

> * if we are going to import `_create_web_driver_at_mint_com`, it isn't really internal and we may as well just remove the beginning underscore.

I recognize that as a common way to divide between public and private attributes but I don't really understand that reasoning. Why does importing make it not internal? I agree that it is no longer internal to the file but I would still think that it would be considered internal as in "not part of our public API" i.e., we are not committing to the user of `mintapi` that it will behave the same way for the major version. As this refactor shows, the way that we divide code between files is rather arbitrary and is really just for developer convenience to make things easier to read; however, what we designate as public or private does affect the user of the application and is separate from where code is held in files.

mrooney:
@jaron-l got it! Can we just name the module "signin" (vs sign_in) then? For some reason the "sign_in_mod" is really bugging me and I don't want to have to name everything else "*_mod" to feel consistent.

> Why does importing make it not internal?

I mean "internal to the file", so importing does make it inherently not internal any more :) However, if you want the underscore to mean it is internal to mintapi as a whole, that seems fine and I'm good with that as is.

If you don't mind just renaming sign_in_mod -> signin then otherwise it looks awesome and good to go!

mrooney (approved):
Let's do it, thanks!

Issue #364, opened 2021-12-08
Only able to log in manually in Chrome after MFA input


msimoni18:
Having trouble logging in with MFA code using ```mintapi=1.58```.

I copied the "from python" section from the README.md and input my username/password. When I run the code, it prompts me for 6-digit MFA code. It continuously asks for the MFA code. In order to log in successfully, I have to manually click the login button in Chrome *after* entering the MFA code for the first time, and *before* the second prompt appears. If the second prompt appears before I can click the login button, the code never finishes. I've also tried adding a path to ```session_path```, but have the same problem.

Is this a bug, or am I doing something wrong?  I'm new to ```mintapi```, so it's entirely possible I'm missing something.

burkematthew:
@msimoni18 can you add a snippet of your code that you're running?

xinluh:
There is a new version of the 2fa page with different id's, e.g. https://github.com/xinluh/yamintapi/commit/b4546d64dd5adf22d2b9e54c088d14da0fe127e4#diff-29db0a202391c066de62d16a0d5b5cc129dc4fb8fb75c5f759a65fdd0452ceef

pazwicker:
Happening to me too. Running:

```python
mint = mintapi.Mint(
        "myemail",  # Email used to log in to Mint
        "mypassword",  # Your password used to log in to mint
        mfa_method="soft-token",  # Can be 'sms' (default) or 'email'.
        headless=True,  # Whether the chromedriver should work without opening a
        mfa_input_callback=None,
        mfa_token='my token',
        session_path=None,
        wait_for_sync=False,  # do not wait for accounts to sync
        wait_for_sync_timeout=1,  # number of seconds to wait for sync
    )
```

msimoni18:
Here's my code that I directly copied from the README.md.  I only changed the username, password, and tried ```session_path=None``` and ```session_path=mypath```.  

```
import mintapi
mint = mintapi.Mint(
    'myusername',  # Email used to log in to Mint
    'mypassword',  # Your password used to log in to mint
    # Optional parameters
    mfa_method='sms',  # See MFA Methods section
                       # Can be 'sms' (default), 'email', or 'soft-token'.
                       # if mintapi detects an MFA request, it will trigger the requested method
                       # and prompt on the command line.
    mfa_input_callback=None,  # see MFA Methods section
                              # can be used with any mfa_method
                              # A callback accepting a single argument (the prompt)
                              # which returns the user-inputted 2FA code. By default
[...]

PR #367, opened 2021-12-09
Adapt to new css selector for submitting OTP

tsdg112:
Mint has a new css selector for submitting a one time password.  This pull request works for either the old or new ways to submit the OTP.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -497,7 +497,7 @@ def sign_in(
                 account_input.click()
 
             mfa_code_submit = driver.find_element_by_css_selector(
-               '#ius-sign-in-mfa-select-account-continue-btn, [data-testid="SelectAccountContinueButton"]'
+                '#ius-sign-in-mfa-select-account-continue-btn, [data-testid="SelectAccountContinueButton"]'
             )
             continue_btn.click()
 
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -474,7 +474,9 @@ def sign_in(
                         )
                     mfa_code_input.send_keys(mfa_code)
 
-                    mfa_code_submit = driver.find_element_by_css_selector('#ius-mfa-otp-submit-btn, [data-testid="VerifyOtpSubmitButton"]')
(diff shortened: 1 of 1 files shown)


msimoni18:
I tested the changes in this PR by cloning the repo, ```python -m pip install -e .```, and then running the code I submitted in #364. It got further this time, but now there is some other error that popped up quickly on Mint saying something went wrong and took me back to the login screen.

pazwicker:
I tested this too and found that we also need to account for a change to `mfa_code_input`:

```python
mfa_code_input = driver.find_element_by_id("iux-mfa-soft-token-verification-code")
```

I got this to work, but I can't for the life of me get the driver to click on "continue." I tried getting by CSS Selector, XPath, link text, but no dice.

mintapi/api.py
```diff
@@ -474,9 +474,15 @@ def sign_in(
...
-                    mfa_code_submit = driver.find_element_by_id(
-                        "ius-mfa-otp-submit-btn"
-                    )
+                    try:
+                        mfa_code_submit = driver.find_element_by_id(
+                            "ius-mfa-otp-submit-btn"
+                        )
+                    except NoSuchElementException:
+                        mfa_code_submit = driver.find_element_by_css_selector(
+                            '[data-testid="VerifyOtpSubmitButton"]'
```
mrooney, line 483: Does find_element_by_css_selector support commas? If so, could we simply do `driver.find_element_by_css_selector('#ius-mfa-otp-submit-btn, [data-testid="VerifyOtpSubmitButton"]')`

tsdg112:
I didn't know that you could do that!  I made the change there and in another place in the code.

mrooney:
I wasn't sure either, but if it is working on your end after the change, I guess we can :D

tsdg112:
Yes, I tested it and its working, at least until Mint makes more changes on their end.

mrooney (approved):
Thanks, looks great to me!

mrooney:
Created #372 to address the potential variable typo.

PR #369, opened 2021-12-09
release/deploy improvements for pypi

mrooney:
This adds a deploy script, which is mostly just convenience for me to perform new uploads to pypi.

It also reverts the long description change since that was running into:

```
The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.
```

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+set -ex
+rm -rf dist && python setup.py sdist
+python -m twine upload dist/*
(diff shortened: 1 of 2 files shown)


mrooney:
@jaron-l happy to explore other fixes to that long_description, I just changed to this temporarily to get the release out. I did the release locally so if we want to change to something else in this PR, that works too.

jaron-l:
Ah, I knew I forgot something. Sorry. Try the recent commit. Should work now.

mrooney:
I like it, let's merge and try in the next release!

PR #370, opened 2021-12-11
fix soft token login

pazwicker:
The ID for soft the soft token input changed on Mint as well as the ID for "Continue." I updated the soft token ID and added try/except for the new continue button.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -421,9 +421,9 @@ def sign_in(
                     mfa_token_submit = driver.find_element_by_css_selector(
                         '#ius-mfa-soft-token-submit-btn, [data-testid="VerifySoftTokenSubmitButton"]'
                     )
+                    mfa_token_submit.click()
                 except NoSuchElementException:
                     pass  # no option to select mfa token
-                mfa_token_submit.click()
 
             else:
                 try:
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -408,7 +408,8 @@ def sign_in(
         try:
             if mfa_method == "soft-token":
                 mfa_token_input = driver.find_element_by_css_selector(
-                    '#iux-mfa-soft-token-verification-code, #ius-mfa-soft-token')
+                    "#iux-mfa-soft-token-verification-code, #ius-mfa-soft-token"
+                )
                 if mfa_input_callback is not None:
                     mfa_code = mfa_input_callback(
                         "Please enter your 6-digit MFA code: "
@@ -421,7 +422,7 @@ def sign_in(
                         '#ius-mfa-soft-token-submit-btn, [data-testid="VerifySoftTokenSubmitButton"]'
                     )
                 except NoSuchElementException:
-                    pass # no option to select mfa token
+                    pass  # no option to select mfa token
                 mfa_token_submit.click()
 
             else:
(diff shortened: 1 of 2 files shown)


mrooney:
Thanks! One thought: sometimes these flows are rolled out to certain or a percentage of users (as I guess you've handled in the second case). So, how do we know no one else is still getting `#ius-mfa-soft-token`? Generally, we try to handle both the old and new cases to account for this. I wonder if we could use `find_element_by_css_selector("#id1,id2")` for a compact use of this? I'm not sure on that syntax, though.

pazwicker:
Just made updates to this PR to look for comma-separated css selectors instead of try/except logic. Also removed xpath reference.

mrooney:
Looks like we had two PRs in flight for this. Does #367 address this same issue, or do we need this as well? Are you able to try with the latest `main` and see if it works for you?

pazwicker:
@mrooney This adds on to #367. Seems like that PR was working on a different stage of the MFA process.

horacix:
I can confirm this PR solves the issue for me and #367 didn't.

mintapi/api.py
```diff
@@ -407,18 +407,24 @@ def sign_in(
...
-                mfa_token_submit = driver.find_element_by_id(
-                    "ius-mfa-soft-token-submit-btn"
-                )
+                try:
+                    mfa_token_submit = driver.find_element_by_css_selector(
+                        '#ius-mfa-soft-token-submit-btn, [data-testid="VerifySoftTokenSubmitButton"]'
+                    )
+                except NoSuchElementException:
+                    pass  # no option to select mfa token
                 mfa_token_submit.click()
```
mrooney, line 426: How would we click this if there's an exception? Does this need to be inside the try? And should we be passing if there's no element, or is that a sign something went wrong and we need to bail and display a failure message?

mrooney:
> @mrooney This adds on to #367. Seems like that PR was working on a different stage of the MFA process.

Got it, will review, thanks!

mintapi/api.py
pazwicker: @mrooney good call. I moved the click inside the try block. But as far as erroring, we'd need to build that logic in for this whole MFA section. Right now it'll pass if there's an error if I'm reading lines 491 and 492 correctly:

```python
except NoSuchElementException:
pass  # not on mfa screen
```

mrooney: Yeah, I guess that was part of my question. If an element wasn't found it would error on trying to click it, and now this changes it to swallow that error more silently. It's fine if there's a good reason I'm not understanding, but if not, we probably shouldn't change the behavior, right?

PS - thanks for this PR and looking forward to merging it!

pazwicker: Hey @mrooney, I'm not confident in my error-handling Python abilities. I'd welcome collaboration here or addressing this error handling in a subsequent PR. I just don't think I'm your guy to write in the right error handling.

mrooney: That's fair, I was just asking why error handling is being *added* in this PR that didn't exist before. So in this case I'd prefer it be removed, not added :) Do you mind just removing the try/except block that was added around lines 420, so the diff is just adding the new selectors, and I'll merge it in? Thanks again for your work!

pazwicker: @mrooney ah, I misunderstood. This is done!

mrooney: Awesome thanks! 🥳

mrooney (approved):

Issue #88, opened 2017-01-08
Add support for exporting accounts to CSV


felciano:
When trying to export account details into an Excel-compatible CSV format I get a ValueError

```
$ mintapi $MINTUSERNAME --extended-accounts --include-investment -f account-details.csv
Traceback (most recent call last):
  ... details omitted ...
    raise ValueError('file type must be json for non-transaction data')
ValueError: file type must be json for non-transaction data
```
I though this used to work and it isn't clear to me why it wouldn't. Can someone confirm? 

Workaround is to install something like https://csvkit.readthedocs.io/, exporting a json version, and then using `in2csv`, but this seems like it would be nice parsimonious functionality to have built in.

mrooney:
It doesn't look like this ever would have worked, I can't recall any time that a format was removed from https://github.com/mrooney/mintapi/blob/master/mintapi/api.py#L750. It looks like we support exporting transactions to csv, but not accounts. I would definitely welcome a PR that abstracts some of the duplicated logic there and supports CSV+JSON for both.

felciano:
So it looks like the current CSV functionality is just downloading Mint's CSV as is. Since you are using pandas for the json data with extended transactions, it should be straightforward to [convert to CSV](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.to_csv.html). 

Can you clarify what CSV compatibility behavior you'd like? This is more of a design question: even if the currently generated CSVs (assuming we use the above pandas function) product results that are identical, Mint CSV export behavior could change such that this no longer holds. So what is the API contract you want to make?

1. Same as what you get from Mint by hand.
2. CSV version of what you get from JSON functions.
3. Both / developer choice, e.g. by add a new function and command line option that produces a CSV dialect guaranteed to match the JSON version.

I sort of like the power of #3 but I'm not sure how on the command line you'd specify which CSV dialect you want.

mrooney:
I think it makes sense, for consistency to be the CSV version of what you get from JSON functions, which I believe is what we do for transaction data: https://github.com/mrooney/mintapi/blob/master/mintapi/api.py#L744 (just call to_csv). So my suggestion would be to just abstract that logic into a function that dates data and the format and handles those options. If that won't handle your use-case though, we can figure out something better!

felciano:
[...]

PR #371, opened 2021-12-12
🧹 Validate file extension before data is retrieved

burkematthew:
This PR includes the following:
* Create a method to perform output filename extensions _before_ we retrieve data.
* Create a method to output the data
* Add test cases for validating that the new filename extension validator works.
* Addresses one of the concerns raised in #88

### DIFF
--- a/mintapi/cli.py
+++ b/mintapi/cli.py
@@ -318,6 +318,54 @@ def handle_password(type, prompt, email, password, use_keyring=False):
     return password
 
 
+def validate_file_extensions(options):
+    if any(
+        [
+            options.transactions,
+            options.extended_transactions,
+        ]
+    ):
+        if not any(
+            [
+                options.filename is None,
+                options.filename.endswith(".csv"),
+                options.filename.endswith(".json"),
+            ]
+        ):
+            raise ValueError(
+                "File extension must be either .csv or .json for transaction data"
+            )
+    else:
+        if not any([options.filename is None, options.filename.endswith(".json")]):
+            raise ValueError("File extension must be .json for non-transaction data")
+
+
+def output_data(options, data, attention_msg=None):
+    # output the data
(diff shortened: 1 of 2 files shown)


mrooney (changes requested):
Seems like a helpful approach to the user and I like the abstraction of the output_data function and test coverage :)

One request: unless I'm missing something, could we just use TemporaryFile instead of NamedTemporaryFile? If we don't need a specific location on disk, it seems potentially more portable, since we are returning the object for writing anyway and not grabbing it by location later.

burkematthew:
> Seems like a helpful approach to the user and I like the abstraction of the output_data function and test coverage :)
> 
> One request: unless I'm missing something, could we just use TemporaryFile instead of NamedTemporaryFile? If we don't need a specific location on disk, it seems potentially more portable, since we are returning the object for writing anyway and not grabbing it by location later.

@mrooney we're using NamedTemporaryFile since we're building a config file to then pass along to parse our command line options, including that of the filename. I ran into some issues with TemporaryFile persisting when implementing the config file test awhile back. If you want me to push through and keep trying, I certainly can. Just let me know!

mrooney:
> > Seems like a helpful approach to the user and I like the abstraction of the output_data function and test coverage :)
> > One request: unless I'm missing something, could we just use TemporaryFile instead of NamedTemporaryFile? If we don't need a specific location on disk, it seems potentially more portable, since we are returning the object for writing anyway and not grabbing it by location later.
> 
> @mrooney we're using NamedTemporaryFile since we're building a config file to then pass along to parse our command line options, including that of the filename. I ran into some issues with TemporaryFile persisting when implementing the config file test awhile back. If you want me to push through and keep trying, I certainly can. Just let me know!

No worries, in that case I'm good, I just wanted to make sure there was a good reason :)

mrooney (approved):
Thanks!

PR #372, opened 2021-12-15
fix soft token login variable name

mrooney:
Unless I'm missing something @tsdg112 , this was perhaps intended in #367 . Let me know if so :)


tsdg112:
That's what I would have done.  Thank you for fixing it.

mrooney:
> That's what I would have done. Thank you for fixing it.

No worries, thanks for verifying! :)

Issue #373, opened 2021-12-15
ci: python 3.10 builds take ~10 minutes


mrooney:
Just opening this to track that python 3.10 builds take ~10 minutes, perhaps caused by building pandas? The other version take ~30 seconds, so this really slows down our cycle/feedback times. Curious if anyone has any ideas!

jaron-l:
It's because the pandas version we have pinned (1.1.4) does not have a wheel version for 3.10. We should simply update the requirements file to something more recent with a 3.10 wheel.

mrooney:
> It's because the pandas version we have pinned (1.1.4) does not have a wheel version for 3.10. We should simply update the requirements file to something more recent with a 3.10 wheel.

Ah, thanks for the pointer! Created #374 to address this :)

mrooney:
#374 merged!

PR #374, opened 2021-12-15
update pandas to latest for python 3.10 wheel and faster builds

mrooney:
Addresses #373. I had to remove python 3.6 (which is EOL in 8 days) as pypi doesn't have a build for a version of pandas that also has a wheel for 3.10. If there's a way to handle both that I missed, let me know!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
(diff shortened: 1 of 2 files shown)


mrooney:
If approved, I'll remove the required check for 3.6 :)

jaron-l (approved):
I confirmed that there is no wheel version of pandas compatible with both python 3.6 and 3.10. Since this really only affects our CI and development and hasn't yet been put into the setup.py, I don't really see a downside in this except we'll stop checking python 3.6. If someone raises an issue, we should upgrade the setup.py python_requires.

mrooney:
Thanks! I removed 3.6 from required checks, and also added 3.10 and black to required checks :)

PR #375, opened 2021-12-18
release version 1.60 with login fixes

mrooney:
This should handle the recent login flow issues as well as recent improvements to categories and such :)

The main reason for the PR is to double check releasing them is good and that I'm not missing anything!


burkematthew:
@mrooney I think because this one has a lot of sign in improvements, it'd make sense to also include #366 .

mrooney:
> @mrooney I think because this one has a lot of sign in improvements, it'd make sense to also include #366 .

Sure thing, it does now, updated!

burkematthew (approved):

jaron-l (approved):
I tested and it works for MFA soft token

PR #377, opened 2021-12-20
🧹 Better handling of exceptions in `get_email_code` 

### LINKED ISSUE #362: bug: :bug: get_email_code hides exceptions when they happen
The convention used throughout the function `get_email_code` to just return an empty string if there is an error hides excpetions. I would suggest that we explicitly raise exceptions when errors happen else the sign workflow will just ignore the problem and fail later when signing in. I think that makes debugging issues harder.

_Originally posted by @jaron-l in https://github.com/mintapi/mintapi/pull/360#discussion_r763277466_

burkematthew:
* Refactored the `sign_in` method for a more modular approach
* Changing `logger.error` instances to exceptions in `get_email_code`

Resolves #362


mintapi/signIn.py
```diff
@@ -46,27 +46,23 @@ def get_email_code(
...
         if rv != "OK":
-            logger.error("ERROR: Unable to open mailbox ", rv)
-            return ""
+            raise RuntimeError("Unable to open mailbox: " + rv)
 
         rv, data = imap_client.search(None, "ALL")
         if rv != "OK":
-            logger.error("ERROR: Email search failed")
-            return ""
+            raise RuntimeError("Unable to search the Email folder")
```
mrooney, line 65: Sometimes we add `rv` to the error message and sometimes not. Would it make sense to always include it like on line 61? I realize you are just converting them as-is; I'm just curious if it can be improved even a little more :)

mintapi/signIn.py
```diff
@@ -75,8 +71,7 @@ def get_email_code(
                 break
             rv, data = imap_client.fetch(num, "(RFC822)")
             if rv != "OK":
-                logger.error("ERROR: ERROR getting message", num)
-                sys.exit(1)
+                raise RuntimeError("Unable to complete due to error message " + num)
```
mrooney, line 74: Is `num` actually meaningful to the user? Would `rv` be better here? I'm not actually sure what it looks like when it isn't "OK" or if `data` is better.

mrooney:
Nice! Does this also handle the issue of making the login flow more dynamic based on what is being prompted for?

burkematthew:
@mrooney no, not yet, though it does make that project clearer. I wanted to get this done first before I made the flow more dynamic.

mintapi/signIn.py
burkematthew: That's a good question. `rv` is most likely the most meaningful and we can standardize the inclusion of that in each message.

mintapi/signIn.py
burkematthew: Yup, I can do that.

burkematthew:
@mrooney back over to you!  Once this merges, I'll work on the more dynamic sign-in flow.

mrooney:
@burkematthew okay, I was having trouble following the diff, but I think I am understanding it better on a second read. So those changes aren't related to the PR title directly, but is basically breaking up one super long function into little parts, that are then called in the same order as before? And in the future, we can add more detection to decide when to call each one?

Are there any potential regressions you could imagine with this? It seems hard to test manually because we can't necessarily control which flow/UI we get 100%, but were you able to test it with the various switches we can provide?

burkematthew:
@mrooney yes I know it made the main change a little more unclear, but I thought it helped understand where exceptions could arise in the overall login flow. 

I will go through some more testing iterations to verify, but it all worked without issue when I was working on this. And really there shouldn't be any regressions unless I just copied/compartmentalized the flow wrong. Again, I'll go through a few more times to verify it's all as expected.

mrooney (dismissed):
Got it, thanks for the clarification! In that case, +1 from me once you are done any final checks you want to do!

burkematthew:
@mrooney all the tests I ran looked good; no issues.

@jaron-l anything else you want to add here?

burkematthew:
FYI, once we get this merged, I'll go ahead and start on the more dynamic MFA handling.

mrooney (approved):

mrooney:
@burkematthew got it, LGTM, approved and merged! I want to unblock you but if @jaron-l has any concerns please do bring them up and we can address them in the next PR :)

Issue #378, opened 2021-12-21
Requesting CSV file for investments per cobolisalive


shumm92:
Only JSON is available currently.

shumm92:
Matthew thank you. I think many users would also appreciate formatting the data set using pandas.

burkematthew:
@shumm92 FYI, you can track progress in #382 .

burkematthew:
@shumm92 this was just merged. If you are running off of main, it's available. Otherwise, it'll be in the next release.

shumm92:
Wow that was fast! Thank you

PR #382, opened 2021-12-22
✨ Feature: Export Investment Data to CSV

burkematthew:
Addresses #378 

* Add `options.investments` to the valid extension list for CSV and JSON
* Add a new method called `json_to_csv` that converts data from a JSON formatted string to a file in CSV format
* Implemented an instance of using `json_to_csv`.

### DIFF
--- a/mintapi/cli.py
+++ b/mintapi/cli.py
@@ -363,6 +363,9 @@ def output_data(options, data, attention_msg=None):
     else:
         if options.filename is None:
             print(json.dumps(data, indent=2))
+        # NOTE: While this logic is here, unless validate_file_extensions
+        #       allows for other data types to export to CSV, this will
+        #       only include investment data.
         elif options.filename.endswith(".csv"):
             json_to_csv(data, options.filename)
         elif options.filename.endswith(".json"):
--- a/mintapi/cli.py
+++ b/mintapi/cli.py
@@ -3,7 +3,6 @@ import logging
 import os
 import sys
 import json
-from csv import writer
 from datetime import datetime
 import getpass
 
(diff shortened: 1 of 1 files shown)


burkematthew:
@jaron-l let me know your thoughts on this one.  Also, I wasn't sure what would be the preferred approach to a test of the new method.  So I'm open to your thoughts there.

mintapi/cli.py
```diff
@@ -375,6 +382,18 @@ def output_data(options, data, attention_msg=None):
                 f.write(attention_msg)
 
 
+def json_to_csv(data, filename):
```
mrooney, line 385: Does this only work for specifically formatted "flat" JSON values? If so, I wonder if there should be some exception handling and a useful error message. If not, it's probably fine!

burkematthew: @mrooney it is, mainly because this method is only used right now for investment data which is flat. If we end up needing to expand to convert other data sources that may not be flat, we'd probably want to have a flattening routine like we have with credit utilization. 

So I could add a comment calling attention to this fact if that would make it clearer and safer?

mrooney: Yeah, my main concern was that the method name implies you could use it to convert any JSON to CSV. So my only nitpick would be that we should either have a check for flatness / compatibility in there that raises a useful error if it fails, and//or rename the method to something like flat_json_to_csv.

burkematthew:
@mrooney I found a better way to go about this by importing `json_normalize` from pandas.  I also left a comment indicating it's only for flat JSON.  Ready for your review again!

mrooney (approved):
Looks very clean, thanks!

PR #384, opened 2021-12-23
Include investment transactions for csv export

### LINKED ISSUE #383: get_transactions and get_transactions_csv no longer uses include_investment
## Describe the bug

When calling the `get_transactions method`: https://github.com/mintapi/mintapi/blob/60519f26b85dce25adff7781ddaa94173461ca8c/mintapi/api.py#L549-L552
the `include_investment` flag is passed. However, the backing `get_transactions_csv` no longer relies on the flag -- it relies on acct instead: https://github.com/mintapi/mintapi/blob/60519f26b85dce25adff7781ddaa94173461ca8c/mintapi/api.py#L512-L517

Right now, this flag can never be 0, and it is impossible to get investment transactions through that method.

## Fix
The flag should be set to 0 if `include_investments == True`

brikeohane:
Fixes #383, mimicking behavior for `get_transactions_json`: https://github.com/mintapi/mintapi/blob/1ea57748ff891d55ba1e7275a60d1214a67622c5/mintapi/api.py#L429

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -835,7 +835,7 @@ class Mint(object):
         return utilization
     
     def __include_investments_with_transactions(id, include_investment):
-        return (id > 0 or include_investment)
+        return id > 0 or include_investment
 
 
 def get_accounts(email, password, get_detail=False):
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -835,7 +835,7 @@ class Mint(object):
                     }
                 )
         return utilization
- 
+
     def __include_investments_with_transactions(id, include_investment):
         return id > 0 or include_investment
 
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -833,9 +833,10 @@ class Mint(object):
                     }
                 )
         return utilization
+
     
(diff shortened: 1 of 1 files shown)


mrooney:
Thanks for your contribution! Requesting a review from @burkematthew as I believe he is recently familiar with this area of code, but please correct me if I'm wrong :)

burkematthew:
@mrooney yes I'll review tomorrow and leave feedback!

burkematthew (changes requested):
@brikeohane thanks for your contribution!  While this would indeed resolve the solution, I'd like to request a standalone method that essentially serves as the control here.  

In `get_transactions_json`, we have `if id > 0 or include_investment:`.  In this code change, you are essentially building the same type of if-statement but within a different source for `> 0`.  I'd like the standalone method to handle the `if ... or ` and then we just check if the method is true or false.

burkematthew:
@brikeohane just sending a follow-up about the above comment.  Thanks!

burkematthew:
Bump @brikeohane

burkematthew (approved):

Issue #385, opened 2021-12-26
CLI throws exception when filename option isn't specified


atkaiser:
**Describe the bug**
As the title states I ran into the exception below when I tried to use the CLI.

What version of MintAPI are you using? 
`1.61`
What command[s] did you run / steps to reproduce?
`python mintapi/api.py <my email>`
Stacktrace/error received:
```
mintapi % python api.py <my email>
/Users/akaiser/Documents/workspace/mintapi/mintapi/api.py:870: DeprecationWarning: Calling command line code from api.py will be deprecated in a future release.
Please call mintapi directly. For examples, see the README.md
  warnings.warn(
Traceback (most recent call last):
  File "/Users/akaiser/Documents/workspace/mintapi/mintapi/api.py", line 877, in <module>
    main()
  File "/Users/akaiser/Documents/workspace/mintapi/mintapi/cli.py", line 388, in main
    validate_file_extensions(options)
  File "/Users/akaiser/Documents/workspace/mintapi/mintapi/cli.py", line 348, in validate_file_extensions
    if not any([options.filename is None, options.filename.endswith(".json")]):
AttributeError: 'NoneType' object has no attribute 'endswith'
```

**What did you expect to happen?**
No exceptions and a JSON  of the output

**What actually happened?**
I got the above exception

**Additional context**
It seems the problem is because inside both the `any` blocks in `validate_file_extensions` a list is being passed in and I guess python needs to evaluate everything in that list before the `any` block is run so if `options.filename is None` then when it tries to evaluate the other things in the list an exception is thrown.  I plan to submit a pull request shortly.

mike503:
this seems to still be happening to me.

ubuntu 20.04
mintapi 1.61 (from pip)
Python 3.8.10

running mintapi -h works.

mintapi with nothing gives this:

```
Traceback (most recent call last):
  File "/usr/local/bin/mintapi", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/dist-packages/mintapi/cli.py", line 388, in main
    validate_file_extensions(options)
  File "/usr/local/lib/python3.8/dist-packages/mintapi/cli.py", line 348, in validate_file_extensions
    if not any([options.filename is None, options.filename.endswith(".json")]):
```

when using mintapi -f something.json it does "run" but I can't actually see it doing anything so far.

however, the expected behavior would be NOT having it throw the error, and also I cannot run any of the documented examples if it doesn't include a -f with some sort of .json as the filename.

[...]

PR #386, opened 2021-12-26
Fixing cli checking file extensions

atkaiser:
Proposed fix for #385 by just removing the `any` and using `or` for the booleans instead.

### DIFF
--- a/mintapi/cli.py
+++ b/mintapi/cli.py
@@ -334,18 +334,16 @@ def validate_file_extensions(options):
             options.extended_transactions,
         ]
     ):
-        if not any(
-            [
-                options.filename is None,
-                options.filename.endswith(".csv"),
-                options.filename.endswith(".json"),
-            ]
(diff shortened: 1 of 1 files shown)


mrooney:
Thanks! I wonder why the `any` wouldn't work, I would think that is functionally the same. Just adding @burkematthew since he wrote this to see if this seems like the best fix.

burkematthew:
@mrooney @atkaiser 

I'll take a look at this tomorrow and see if I can figure out what's going on.

jaron-l:
They're not the same because the `or` allows the following tests to not be evaluated if `filename` is none. They code fails as is because when `filename` is none, it doesn't have the endswith attribute.

jaron-l (approved):

mrooney:
@jaron-l thanks for the review and merge, @atkaiser thanks for the PR!

Issue #388, opened 2021-12-27
Uncaught Selenium Exception: StaleElementReferenceException


jeremyehack:
**Describe the bug**
I'm running mintapi in a container headless on AWS and it has been working fine for months even through occasional Chrome and package version updates. It runs through the login flow fine, but then after it enters MFA it is erroring on a StaleElementReferenceException. I ran same script without headless mode on my Mac and was able to confirm that it is happening right after the MFA soft-token flow completes. 

What version of MintAPI are you using?
  * mintapi==1.60

What command[s] did you run / steps to reproduce?
Tried a couple ways:
  * Building driver through Chrome Opts:
    ```
        mint = mintapi.Mint()
        mint.driver = Chrome(executable_path=f'/opt/chromedriver', chrome_options=chrome_opts())
        mint.status_message, mint.token = mintapi.sign_in(email, password, driver=mint.driver, mfa_method="soft-token", mfa_token=token)
    ```
  * Using driver on path:
    `mint = mintapi.Mint(email=email, password=password, use_chromedriver_on_path=True, headless=True, mfa_method="soft-token", mfa_token=token)` 

Stacktrace/error received:
```
  (Session info: headless chrome=96.0.4664.110)

Traceback (most recent call last):
  File "/var/task/mintapi/api.py", line 235, in login_and_get_token
    self.status_message, self.token = sign_in(
  File "/var/task/mintapi/signIn.py", line 482, in sign_in
    driver.find_element_by_id(
  File "/var/task/selenium/webdriver/remote/webelement.py", line 477, in send_keys
    self._execute(Command.SEND_KEYS_TO_ELEMENT,
  File "/var/task/selenium/webdriver/remote/webelement.py", line 633, in _execute
    return self._parent.execute(command, params)
  File "/var/task/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/var/task/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.StaleElementReferenceException: Message: stale element reference: element is not attached to the page document
```

**What did you expect to happen?**
Successfully login and the Mint object gets data that it is looking for

**What actually happened?**
Getting above error when using the mfa soft-token flow. 

**Additional context**
[...]

PR #391, opened 2021-12-31
don't fail sign-in on a StaleElementException

mrooney:
As per #388, catching this and passing has allowed various folks to continue onwards on their sign-in journey.


jaron-l (approved):

burkematthew:
Note that this fixed #388 .

PR #398, opened 2022-01-29
🐛 Fixing regression from #384

burkematthew:
I caused a regression in the slight refactoring of PR #384.  This resolves the regression.


mrooney (approved):
Wow, a test to catch it later too, nice!

Issue #387, opened 2021-12-27
get_transactions_json() failure


bfongbfong:
**Describe the bug**
firing `mint.get_transactions_json()` is failing for me. 

What version of MintAPI are you using?
1.61

What command[s] did you run / steps to reproduce?
fire `mint = mintapi.Mint( ... )`, and then run 
`mint.get_transactions_json(include_investment=False, skip_duplicates=True)`
Stacktrace/error received:
```
File "__main__.py", line 199, in fire_mint
    transactions_json = mint.get_transactions_json(include_investment=False, skip_duplicates=True)
  File "/Users/brandonfong/Library/Python/3.8/lib/python/site-packages/mintapi/api.py", line 405, in get_transactions_json
    self.set_user_property("hide_duplicates", "T" if skip_duplicates else "F")
  File "/Users/brandonfong/Library/Python/3.8/lib/python/site-packages/mintapi/api.py", line 376, in set_user_property
    result = self.make_post_request(
  File "/Users/brandonfong/Library/Python/3.8/lib/python/site-packages/mintapi/api.py", line 200, in make_post_request
    response = self.post(url=url, data=data, headers=JSON_HEADER)
  File "/Users/brandonfong/Library/Python/3.8/lib/python/site-packages/mintapi/api.py", line 197, in post
    return self.driver.request("POST", url, **kwargs)
  File "/Users/brandonfong/Library/Python/3.8/lib/python/site-packages/seleniumrequests/request.py", line 163, in request
    window_handles_before = len(self.window_handles)
  File "/Users/brandonfong/Library/Python/3.8/lib/python/site-packages/selenium/webdriver/remote/webdriver.py", line 724, in window_handles
    return self.execute(Command.W3C_GET_WINDOW_HANDLES)['value']
  File "/Users/brandonfong/Library/Python/3.8/lib/python/site-packages/selenium/webdriver/remote/webdriver.py", line 319, in execute
    response = self.command_executor.execute(driver_command, params)
  File "/Users/brandonfong/Library/Python/3.8/lib/python/site-packages/selenium/webdriver/remote/remote_connection.py", line 374, in execute
    return self._request(command_info[0], url, body=data)
  File "/Users/brandonfong/Library/Python/3.8/lib/python/site-packages/selenium/webdriver/remote/remote_connection.py", line 397, in _request
    resp = self._conn.request(method, url, body=body, headers=headers)
[...]

PR #401, opened 2022-02-19
Re-throw exception during login

zsolteszku:
By silencing the exception in the `login_and_get_token` the code execution could still continue without an explicit try-except which would could end up additional calls to a `WebDriver` that was already closed (by `quit()`) that leads to confusing error messages (similar to `MaxRetryError("HTTPConnectionPool(host='127.0.0.1', port=55873): Max retries exceeded with url: /session/2dd60780e17e5d8d8e1cdf9496086729/window_handles`)

Could be related to issue: #387

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -248,8 +248,11 @@ class Mint(object):
                 imap_folder,
             )
         except Exception as e:
+            msg = f"Could not sign in to Mint. Current page: {self.driver.current_url}"
             logger.exception(e)
             self.driver.quit()
+            self.driver = None
+            raise Exception(msg) from e
 
     def get_request_id_str(self):
         req_id = self.request_id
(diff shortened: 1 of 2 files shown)


burkematthew:
@zsolteszku we are redoing a lot of login flow in the PR#392. Can you check out this branch and see if your issue persists? Also, I'd greatly appreciate any testing feedback you could provide so I could get more perspectives of login flow.

zsolteszku:
@burkematthew I tested the new branch and this issue still persist there as well as the exception is not re-thrown.

burkematthew:
@zsolteszku thank you for testing that out! I'll review your change.

burkematthew (approved):

mrooney:
Amazing, thanks for the fix with test! 👯

PR #402, opened 2022-02-19
Able to set the WebDriver to use in constructor

zsolteszku:
Usecase: Anyone could pass their own custom webdriver to the Mint. This will enable users to run the script on AWS Lambda (where you would need to have a special headless chrome to be able to run) or any container where a customer driver is necessary.


mrooney:
Looks pretty straightforward, thanks! Can you update the PR and ideally README / docs with the use-case this addresses?

mrooney (changes requested):
Looks pretty straightforward, thanks! Can you update the PR and ideally README / docs with the use-case this addresses?

zsolteszku:
Usecase: Anyone could pass their own custom webdriver to the Mint. This will enable users to run the script on AWS Lambda (where you would need to have a special headless chrome to be able to run) or any container where a customer driver is necessary.

@mrooney Would you like me to include this usecase in the README?

burkematthew:
@zsolteszku yes a README update would be great, thanks!

mrooney:
@zsolteszku Agreed, that would be fantastic, especially if it had an example for how to do this on AWS with the appropriate driver :)

zsolteszku:
@burkematthew and @mrooney updated the README.md, let me know your thoughts.

burkematthew (approved):

mrooney (approved):

PR #414, opened 2022-03-07
Update overview page url

hudcap:
For me, the overview page no longer has ".event" at the end, so mintapi hangs here.

### DIFF
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -323,7 +323,7 @@ def sign_in(
 
     # Wait until logged in, just in case we need to deal with MFA.
     driver.implicitly_wait(1)  # seconds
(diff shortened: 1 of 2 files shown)


mrooney:
LGTM, since this is a `startswith` it should still handle the old URL if that still appears for some folks. Did you verify that this fixes the issue for you?

hudcap:
Yes, although I then hit some other snag with the "javascript-user" that I haven't had a chance to investigate yet.

burkematthew:
@hudcap yes the `javascript-user` issue was what I was investigating this morning. 

@mrooney I say we merge this and then I can circle back around and take care of the other unrelated issue as mentioned above.

burkematthew (approved):

PR #420, opened 2022-03-12
[FIX] - Use the new location of the Mint API Key

burkematthew:
This is the next part of a series of fixes required based on the new layout and flow of Mint.  In the update, their scripts changed and so the location of the API Key changed.  

NOTE: This doesn't completely fix the problems that users are experiencing, but is one necessary step in the process.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -144,7 +144,7 @@ class Mint(object):
         ).zfill(3)
 
     def _get_api_key_header(self):
-        key_var = "window.MintConfig.browserAuthAPIKey"
+        key_var = "__shellInternal.OILConfigs.key"
         api_key = self.driver.execute_script("return " + key_var)
         auth = "Intuit_APIKey intuit_apikey=" + api_key
         auth += ", intuit_apikey_version=1.0"


mintapi/api.py
```diff
@@ -144,7 +144,7 @@ def get_rnd(cls):  # {{{
         ).zfill(3)
 
     def _get_api_key_header(self):
-        key_var = "window.MintConfig.browserAuthAPIKey"
+        key_var = "__shellInternal.OILConfigs.key"
```
mrooney, line 147: That is really something!

mrooney (approved):

tsdg112:
I have several Mint accounts.  Some of them are still using the javascript-user flow and work fine with the current Mintapi.  Will this change break retrieving data from the accounts that have not yet been converted?

mrooney:
> I have several Mint accounts. Some of them are still using the javascript-user flow and work fine with the current Mintapi. Will this change break retrieving data from the accounts that have not yet been converted?

That's a great question! In the ideal world it would be great if the new endpoints worked for all accounts so it didn't matter. Perhaps you can test before a release and we can verify. @burkematthew let me know if you have any guesses one way or the other.

burkematthew:
@mrooney @tsdg112 I'd imagine that they're simply rolling out the UI to phases of users and it's only a matter of time before all users have the new UI. The endpoints should have no problem working for those users even that still have the old UI.

I do agree that it'd be good to have you test before the next release to verify all is working as expected for you on the old flow. In the meantime, we are going to be pushing forward to get the new flow working.

mrooney:
@burkematthew sounds good! I think in a worst case, in the interim we can always say to install the older version from pypi if your account isn't rolled out yet.

PR #421, opened 2022-03-12
[FIX] - Remove deprecated get_token functionality

burkematthew:
This is another part of a series of fixes required based on the new layout and flow of Mint. In the update, we no longer could reference an element called `javascript-user` to pull a token.  All API Requests use the API Key header as authentication so we can simply remove the `get_token` functionality.

NOTE: This doesn't completely fix the problems that users are experiencing, but is one necessary step in the process.


mrooney (approved):
LGTM, thanks!

PR #423, opened 2022-03-19
[FIX] - Use the correct field for the API Token

burkematthew:
PR #420 attempted to return the API Key from the new location.  However, after further testing, I discovered the wrong return value was being used.  This PR uses the correct reference.


mrooney (approved):
Nice find!

PR #425, opened 2022-03-19
🔨 - Update Budgets Endpoint

burkematthew:
This PR addresses the outdated Budgets endpoint with the switch to the new Mint UI.  For v1 of this update, we are going to keep the formatting as is returned from the API, except we are popping out the metadata so that each element can be flattened for a CSV.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -616,75 +616,23 @@ class Mint(object):
 
         return accounts
 
-    def get_budgets(self, hist=None):
-        response = self.__call_budgets_endpoint()
-        categories = self.get_categories()
-        income = response["data"]["income"]
-        spending = response["data"]["spending"]
-        if hist is not None:  # version proofing api
-
-            def mos_to_yrmo(mos_frm_zero):
-                return datetime(
-                    year=int(mos_frm_zero / 12), month=mos_frm_zero % 12 + 1, day=1
-                ).strftime("%Y%m")
-
-            # Error checking 'hist' argument
-            if isinstance(hist, str) or hist > 12:
-                hist = 12  # MINT_ROOT_URL only calls last 12 months of budget data
-            elif hist < 1:
-                hist = 1
-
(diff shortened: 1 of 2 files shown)


burkematthew:
Need to address failing tests.

mintapi/api.py
```diff
@@ -616,75 +616,23 @@ def populate_extended_account_detail(self, accounts):  # {{{
...
-        }
-        return json.loads(self.get(url, params=params, headers=JSON_HEADER).text)
-
-    def __format_budget_categories(self, budget, categories):
-        category = self.get_category_object_from_id(budget["cat"], categories)
-        budget["cat"] = category["name"]
-        parent = self._find_parent_from_category(category, categories)
-        budget["parent"] = parent["name"]
-        return budget
+        return self.get(
```
mrooney, line 630: Is this pattern used elsewhere? I'm curious if we should have some kind get_json_with_headers if so. If not, no worries.

burkematthew: @mrooney Yeah this will be a pattern that will be very prevalent in the new endpoint changes.  Once this merges, I can do some cleanup where I create a generic method to better encompass this functionality.

mintapi/api.py
```diff
@@ -616,75 +616,23 @@ def populate_extended_account_detail(self, accounts):  # {{{
...
-                "spend": spending[str(max(map(int, spending.keys())))]["bu"],
-            }
-
-            # Fill in the return structure
-            for direction in budgets.keys():
-                for budget in budgets[direction]:
-                    budget = self.__format_budget_categories(budget, categories)
-
-        return budgets
+            raise MintException("Cannot find budget data")
```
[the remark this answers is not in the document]
burkematthew: @mrooney Good point.  Since this is consistent with error messaging with Category and Investment so far, I'd say let's leave this as is for now and then do a follow-up PR that handles more specific error handling from the new endpoints.

mrooney (approved):

PR #427, opened 2022-03-20
🔨 - Remove support for hiding duplicates

burkematthew:
After reviewing Mint's new UI, there is no trace of support for hiding duplicate transactions.  As a result, we are removing support for the functionality from Mintapi.


mrooney (approved):
Nice cleanup! Thanks for doing this :)

PR #430, opened 2022-03-27
🔨 - Update Accounts Endpoint

burkematthew:
This PR addresses the outdated Accounts endpoint with the switch to the new Mint UI.  For v1 of this update, we are going to keep the formatting as is returned from the API, except we are popping out the metadata so that each element can be flattened for a CSV.

Note that there is NO LONGER an `extended-accounts` functionality.  This functionality was a nasty workaround that involved scrapping div elements off the screen.  Considering that the majority of the data is already in the response and that the UI changed, Mintapi will no longer attempt to pull data in this fashion.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -1,7 +1,5 @@
-from datetime import date, datetime, timedelta
+from datetime import date, datetime
 from dateutil.relativedelta import relativedelta
-import io
-import json
 import logging
 import os
 import random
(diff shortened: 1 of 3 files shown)


mrooney (dismissed):

mrooney:
Approved although it looks like black ran into an error 🤔

burkematthew:
@mrooney just FYI, the issue we're seeing with black should be fixed when this patch is released: https://github.com/psf/black/pull/2966

burkematthew:
All good now!

chino:
Is there any info on what we'll lose without `--extended-accounts`?  I only ask because I was previously using that and will need to update.

Issue #89, opened 2017-02-02
choose output as csv or json when printing output to stdout


sken77:
when transaction output is sent to `stdout` allow to be able to choose between csv and json

mrooney:
Seems reasonable, PRs welcome!

bradmalloy:
I'll take a crack at this

burkematthew:
@jaron-l I know you had some thoughts about this when discussing #256 .  What are your thoughts on this?

jaron-l:
> @jaron-l I know you had some thoughts about this when discussing #256 . What are your thoughts on this?

My only concern is to avoid breaking the current workflow. So since json is the default, csv could be a flag to change the stdout.

burkematthew:
@jaron-l yes I was thinking that it could be a new option of --csv or something like that to indicate we want to format as CSV for stdout.

burkematthew:
@sken77 when thinking of output as CSV, do you mean a true comma-separate list or a pandas data frame?

burkematthew:
Hi @sken77 just following up here.

burkematthew:
Hi @sken77 just following up here.

PR #432, opened 2022-03-29
[FEATURE] - Add Data Format Option to CLI

burkematthew:
Currently, the format of the output is determined based on the file extension specified.  In addition, the only available format when printing to `stdout` is json.  This PR makes a couple of enhancements:

* A new CLI option is added: `--format`.  Its valid values are `csv` and `json`, with `json` as the default answer.
* The filename no longer includes the extension.  It will be determined based on the value of `--format`.
* If no filename is specified, then the format in `stdout` is determined based on `--format`.

This PR addresses #89 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,21 @@
+2.0 (Pending)
+---
+
+- Add Data Format Option to CLI: `--format` (#432)
+- Update Accounts Endpoint to meet new Mint requirements (#430) 
+- Update Transactions Endpoint to meet new Mint requirements (#429) 
+- Update Budgets Endpoint to meet new Mint requirements (#425)
+- Support removed for hiding duplicate transactions (#427)
+- Fetch the correct API Key to use with Mint API Requests (#420, #423)
+- Removed `get_token` functionality, which is incompatible with the new Mint UI (#421)
+- Update the name of the Account Refresh Class (#415)
+- Update the overview page url (#414) 
+
+BREAKING CHANGES:
+- Because of the new Mint UI and the switchover to different API Endpoints, the data structure associated with each function may be different.  Please verify that the data you are expecting against the new output of each endpoint.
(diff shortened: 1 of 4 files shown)


mrooney:
Nice work! I like that it allows formatting of the non-file output and seems to clean up some areas. One thought: is there a big advantage to writing the filename automatically? I ask because it seems nice not to break existing flows if possible (folks will have to change another argument) and personally I'm a fan of having an explicit filename option versus having something change it for me. It's also nice because you might want to write to a specific location that something else is reading from that isn't looking for an extension (`/tmp/output` or whatever).

mrooney (changes requested):
Could you also start the pending 2.0 changelog in this PR and add a note for the new flag? :)

burkematthew:
@mrooney a couple things:

1. Yes I can add the pending PR log. 
2. We are already requiring certain file extensions on output. If you don't specify csv or json, it errors out already. So this is really just changing it from evaluating the filename value for the extension type to evaluating the option. 
3. I think it could get confusing if we have both the extension and the format option. Then, we have to handle mismatched extension and format. Also, it won't technically break if they include the extension as it will just have the extension twice in the resulting filename. 

We're already introducing a decent bit of breaking changes and I think this one in particular seems a bit trivial comparatively. Unless you really object, I think we should keep it as is designed. 

Also, I'm thinking ahead a little that when we allow multiple data types in one request, we'll want to allow outputting each data type to its own file and then I think the filename functionality may have even further changes to support that.

burkematthew:
@mrooney CHANGELOG updated!

mrooney:
Thanks!

> Then, we have to handle mismatched extension and format.

If someone wants to do --format=json --filename=foo.csv that's their (weird) choice; it should be formatted as json and placed at foo.csv IMHO :P

> Also, it won't technically break if they include the extension as it will just have the extension twice in the resulting filename.

I guess I'm fine with that as long as everything can also be written to stdout the same way, so you can do something like `mintapi ... | /dev/foo`. It seems like that would be the case? I just want to avoid a use-case where you have to use a file output and want it to go to some pipe/device where appending a filename would break it. If anything that can go to a file can go to stdout, it seems still just as flexible/

mrooney (approved):

PR #433, opened 2022-03-29
🔨 - Cleanup unused methods

burkematthew:
This PR does a sweep through the source and removes any unused source.


tests/test_driver.py
```diff
@@ -334,12 +318,6 @@ def test_config_file(self):
         self.assertEqual(arguments.transactions, True)
         config_file.close()
 
-    @patch.object(mintapi.signIn, "get_web_driver")
-    def test_build_bundledServiceController_url(self, mock_driver):
```
mrooney, line 338: What does it mean for a test method to be unused? Isn't this test running?

mrooney (changes requested):
Did you use a tool for this? Would it warrant a CI check if so? Let me know about the test methods and also, aren't some methods like get_web_driver() exposed for end-users?

tests/test_driver.py
burkematthew: The bundled service method is no longer needed and so this test is no longer needed as well.

burkematthew:
@mrooney I removed `get_web_driver` because it was marked as deprecated and thought that was the intent at next major version. And I went through and just manually audited the source code.

mrooney (approved):

PR #434, opened 2022-03-29
🔨 - Dynamically use Endpoints

burkematthew:
This PR creates general methods for getting data from an endpoint using settings configured in a constant dictionary.  The settings include the following settings:

* apiVersion
* endpoint
* beginningDate
* endingDate
* includeCreatedDate

`beginningDate` is used to specify a potential beginning date query parameter in the API request.  If there is none, we set this to `None`.
`endingDate` is used to specify a potential ending date query parameter in the API request.  If there is none, we set this to `None`.
`includeCreatedDate` is used to toggle whether the metadata object of the response includes `createdDate`.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -18,48 +18,43 @@ CATEGORY_KEY = "Category"
 INVESTMENT_KEY = "Investment"
 TRANSACTION_KEY = "Transaction"
 
-ENDPOINTS = [
-    {
-        "key": ACCOUNT_KEY,
+ENDPOINTS = {
+    ACCOUNT_KEY: {
         "apiVersion": "pfm/v1",
         "endpoint": "accounts",
         "beginningDate": None,
         "endingDate": None,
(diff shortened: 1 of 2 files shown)


mintapi/api.py
```diff
@@ -520,6 +516,29 @@ def _flatten_utilization(self, data):
                 )
         return utilization
 
+    def __find_endpoint(self, name):
```
mrooney, line 519: It seems like the endpoints is an array and each one has a "key", and then we have to find it inside. Is there a reason to not make ENDPOINTS a dict with a key so it actually is a key and you can just grab the rest of the values from it?

burkematthew: @mrooney very good point! I can do that.

burkematthew: @mrooney done!

mrooney (approved):
Looks great, thanks!

PR #438, opened 2022-04-01
[FIX] - Correct the documentation for --show-pending

hudcap:


hudcap:
Oops, still new to this git stuff...
I'm gonna retry without those old commits

PR #443, opened 2022-04-02
[FEATURE] - Add parameter to control limit of records

burkematthew:
As was discussed in a Discord thread, the default limit on each endpoint controlled by Mint is not large enough.  This PR addresses a couple items:

* Add support for a limit argument
* Add CLI support for a limit argument
* Set the default limit to 5000 records.

### DIFF
--- a/mintapi/cli.py
+++ b/mintapi/cli.py
@@ -196,6 +196,14 @@ def parse_arguments(args):
             },
         ),
         (
+            ("--limit",),
+            {
+                "type": int,
+                "default": 5000,
+                "help": "Number of records to include from the API.  Default is 5000.",
+            },
+        ),
+        (
             ("--mfa-method",),
             {
                 "choices": ["sms", "email", "soft-token"],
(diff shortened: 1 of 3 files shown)


mintapi/cli.py
```diff
@@ -411,42 +419,47 @@ def main():
...
     elif options.budget_hist:
         try:
-            data = mint.get_budgets(hist=12)
+            data = mint.get_budgets(limit=options.limit, hist=12)
         except Exception:
             data = None
     elif options.accounts:
         try:
-            data = mint.get_account_data()
+            data = mint.get_account_data(options.limit)
```
mrooney, line 444: Should we specify this by keyword, `limit=options.limit`, for consistency? Otherwise looks great!

mrooney (changes requested):

mintapi/cli.py
burkematthew: @mrooney yup, I just missed that one. Thanks! I'll get that fixed up tonight.

burkematthew:
@mrooney back to you!

mrooney (approved):
Nice feature, thanks! :)

PR #448, opened 2022-04-07
🐛 - Use explicit wait for Mint Overview to load

burkematthew:
This PR addresses a couple items:

1. Removing some unused arguments passed to methods
2. Fix the status message returned from waiting for sync
3. The main issue here is that there isn't enough of a pause\delay in waiting for the overview page to load.  Add an explicit wait for 5 seconds to let the page load.

### DIFF
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -11,7 +11,6 @@ import subprocess
 import sys
 import time
 import zipfile
-import warnings
 
 from selenium.common.exceptions import (
     ElementNotInteractableException,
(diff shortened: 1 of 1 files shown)


mintapi/signIn.py
```diff
@@ -369,12 +368,14 @@ def sign_in(
         )
         account_selection_page(driver, intuit_account)
         password_page(driver, password)
+        # Give the overview page a chance to actually load
+        time.sleep(5)
```
mrooney, line 372: Is it possible to use some kind of [explicit wait](https://www.browserstack.com/guide/selenium-wait-for-page-to-load) here or page load event? A sleep doesn't seem super robust as the load could take longer on a slow connection, or it could be done faster and we're just slowing it down.

mrooney (changes requested):

mintapi/signIn.py
burkematthew: @mrooney I'll look at what you linked and see. The reason behind this is we don't know for sure if we're going to the main page or looping through to the next round of MFA.

mrooney (approved):

PR #450, opened 2022-04-09
[FEATURE] - Support multiple data types from CLI

burkematthew:
The CLI has been refactored to support exporting multiple data types in one command.  When exporting multiple data types, you can either send it directly to the `stdout` or you can export to a data file. What MintAPI will do with your specified filename is add a suffix based on the type of data you are exporting. The following table outlines the option selected and its corresponding suffix:

| Option       | Suffix       |
| -----------  | -----------  |
| accounts     | account      |
| budgets      | budget       |
| transactions | transaction  |
| categories   | category     |
| investments  | investment   |
| net-worth    | net_worth    |
| credit-score | credit_score |
| credit-report| credit_report|

For example, if you specify `current` as your filename, format as csv, and you export `account` and `transaction`, then you will receive two files: `current_account.csv` and `current_transaction.csv`.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -1,5 +1,6 @@
 from datetime import date, datetime
 from dateutil.relativedelta import relativedelta
+import mintapi.constants as constants
 import logging
 import os
 import random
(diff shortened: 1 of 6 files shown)


burkematthew:
@mrooney let me know what you think!

mintapi/api.py
```diff
@@ -1,5 +1,6 @@
 from datetime import date, datetime
 from dateutil.relativedelta import relativedelta
+import mintapi.constants as constants
```
mrooney, line 3: Normally I would do `from mintapi import constants` but this does make the imports a little visually cleaner to do it this way. Is there any difference?

mintapi/cli.py
```diff
@@ -416,63 +413,83 @@ def main():
...
-    if options.attention:
-        attention_msg = mint.get_attention()
-    output_data(options, data, attention_msg)
+    if options.categories:
+        try:
+            data = mint.get_categories(
+                limit=options.limit,
+            )
+            output_data(options, data, constants.CATEGORY_KEY, attention_msg)
+        except Exception:
```
mrooney, line 459: Does this make sense to replicate any more? Is setting `data = None` providing any utility, or blindly catching all exceptions? I wonder if we could simply remove these calls from the try/except and tidy it up?

mrooney (changes requested):
Awesome thanks, just a couple questions!

mintapi/api.py
burkematthew: No difference. I'd say just consistency. I think I veered off of convention so I can change it.

mintapi/cli.py
burkematthew: Very valid point. I have a desire to refactor it a little more once it's out in the wild. 

But yeah it's really just blindly catching exceptions and initializing the variable, though probably unnecessary.

tests/test_driver.py
```diff
@@ -5,7 +5,7 @@
 import unittest
 import requests
 import tempfile
-
+import mintapi.constants as constants
```
mrooney, line 8: Nice, looks much cleaner! Do you mind just updating this to the "from mintapi import constants" as well for consistency?

mrooney (changes requested):

mrooney (approved):

felciano:
@mrooney should this excellent merge have also included a version number bump to 2.0, per the README?

PR #451, opened 2022-04-10
[FIX] - Use url_contains instead of url_changed

burkematthew:
As a follow-up to #448 , more testing showed that we should use `url_contains` instead of `url_changed` for managing the explicit wait.


mrooney (approved):
Thanks!

PR #452, opened 2022-04-17
Account for the click of your MFA method

tsdg112:
With email MFA, one first needs to click “Email a code” to start the email MFA.  This PR adds this click.  The current code conflates the ids used for clicking “Email a code” with the ids for entering the code as INPUT_CSS_SELECTORS_LABEL.  I separated out the ids for clicking “Email a code” into SELECT_CSS_SELECTORS_LABEL and added the click.

This PR has only been tested for MFA via email.  I don’t have access to the other authentication types.  Looking at the 1.66 branch, it appears that this initial click is not needed for soft token or authenticator MFA.  Somebody needs to verify that this PR did not break these MFA methods.

### DIFF
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -34,6 +34,7 @@ MFA_VIA_AUTHENTICATOR = "authenticator"
 MFA_VIA_EMAIL = "email"
 MFA_VIA_SMS = "sms"
 MFA_METHOD_LABEL = "mfa_method"
+SELECT_CSS_SELECTORS_LABEL = "select_css_selectors"
 INPUT_CSS_SELECTORS_LABEL = "input_css_selectors"
 SPAN_CSS_SELECTORS_LABEL = "span_css_selectors"
 BUTTON_CSS_SELECTORS_LABEL = "button_css_selectors"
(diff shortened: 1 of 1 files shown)


tsdg112:
This PR does not properly handle the case when it needs to both process the MFA and bypass the verified user page.  The flow in mint is for the bypass page to happen after the MFA.

The 1.66 branch handles this by iterating the “while not driver.current_url.startswith” loop to call bypass_verified_user_page.   But here, we get a timeout exception at the WebDriverWait and the loop never repeats.  I don’t understand why the   WebDriverWait was added, and don’t want to change that code without further guidance.

Also, it’s really difficult to debug the email MFA code because Mint only wants to do it once per account.  Is there any easy way to force Mint to do the MFA on the next login?

burkematthew:
@tsdg112 thanks for the PR and attempting to bring another perspective of input to the dynamic flow we implemented. I'll find some time this week to take a look and review.

tsdg112:
I fixed the "while not driver.current_url.startswith" loop:
I moved the code to enter credentials out of the loop, so that its only done once.
I removed WebDriverWait from the loop and replaced it with a loop counter.

This has been successfully tested with the email MFA.

@burkematthew Please review.

burkematthew:
@tsdg112 I'm not quite sure this will work. The reason it was included is because when Mint asks for multiple MFA methods, which is possible, it prompts for username and password multiple times. 

From what it sounds, the underlying issue is that Mint can ask WHICH MFA method you want to use, requiring a click to make the selection. I think that click handle is what is missing and the loop itself and the username/password in the loop is a non-issue.

tsdg112:
The trouble is that no one developer sees all the different ways to interact with Mint.  In my case, I only have a single MFA method, but I still need to click on "email a code" first.

If I put the code to enter credentials back into the loop and login succeeds in the first iteration, then handle_different_page_username_password with throw an exception in the second iteration.  Would it make sense to put the credentials code back into the loop with a try/except pass block to prevent it?

burkematthew:
@tsdg112 the reason for the wait was because the page needed a minute for the overview page to actually load. Otherwise, as you have observed, the loop goes through a second time.

And also, I agree that no on developer sees all the ways to interact with Mint. That being said, the intent behind the way the code is structured now is to dynamically handle when a user is prompted for an MFA, even if there are multiple MFA methods requested, which is a valid use case. 

Again, really all that seems to be necessary here is implementing the ability to click the MFA method option when presented with that screen.

tsdg112:
I've merged in the changes from #455.  I also fixed an additional problem where it would consistently get "Mint sync apparently incomplete after timeout" after bypassing the verified user page.  This was happening because it would unnecessarily try the MFA again.  That made it take too long to exit the while loop, so that the sync message was gone before handle_wait_for_sync checked for it,

burkematthew:
Thanks @tsdg112 ! Can you fix the merge conflicts? Once that's resolved, I'll do my set of tests and then hopefully should be good to go. Thanks again!

tsdg112:
@burkematthew I'm a github novice and don't understand the conflict or how to fix it.  I think that the implicitly_wait that its complaining about belongs in the code.

burkematthew:
@tsdg112 I'll review it and resolve. If I have any questions, I'll let you know.

mintapi/signIn.py
```diff
@@ -523,6 +540,10 @@ def search_mfa_method(driver):
     for method in MFA_METHODS:
         mfa_token_input = mfa_token_button = mfa_method = span_text = result = None
         try:
+            mfa_token_select = driver.find_element_by_css_selector(
```
burkematthew, line 543: @tsdg112 this section won't work.  This assumes that every user has the "select" page, which they will not.  In fact, this is the core of the "dynamic MFA flow" that we built.

mintapi/signIn.py
```diff
@@ -549,6 +570,10 @@ def set_mfa_method(driver, mfa_method):
     )
     mfa_result = list(mfa)[0]
     try:
+        mfa_token_select = driver.find_element_by_css_selector(
```
burkematthew, line 573: I believe THIS is the correct change needed to do what you're expecting.

The key, though, is if you want to select "email" as the MFA method at the choose screen, you need to pass in `mfa-method=email` as an argument.

burkematthew (changes requested):

mintapi/signIn.py
tsdg112: Thank you for catching this.  I made the change.

burkematthew: @tsdg112 you bet! I'll test this out again this afternoon and make sure we're good. Then I'll be good to go here!

burkematthew:
My tests looked good.  I tried both with a single MFA (email via IMAP) and a multi-MFA (email via IMAP and text message).  Both still completed as expected.  Approved!

burkematthew (approved):

PR #392, opened 2022-01-04
🔨 Dynamic Multifactor Authentication Flow

burkematthew:
Currently, we rely on user-submitted `mfa_method` to match with the configured settings in Mint.  This can lead to a lot of confusion surrounding correct flow and often leads to a bit of troubleshooting when login doesn't work as expected.  The submitted approach does the following:

* A new array, `MFA_METHODS`, is available to control which Multifactor Authentication methods are available and the corresponding CSS elements on the screen.
* Because of the opportunity for Mint to prompt the user to select a Multifactor Authentication method, we are preserving the `mfa_method` argument and pass it to `mfa_selection_page` in the case that the relevant UI Components exist on the screen.
* In the case that `mfa_method` is supplied, we search `MFA_METHODS` for the resulting method and locate the CSS elements that should exist on the screen.
* In the case that `mfa_method` is not supplied, we loop through `MFA_METHODS` looking for matches of the CSS elements on the screen.  Once we find one, we proceed with the key for `mfa_method`, the input element on the screen, and the button element on the screen.
* Lastly, we pass each of those to individual methods based on if an One-Time Password (OTP) is required (`soft-token`), an email onscreen with IMAP configured, or any other type of Multifactor Authentication code.

This should fix #291 and fix #254.

PR #454, opened 2022-04-23
[FIX] - Typo on soft-token constant

burkematthew:
When implementing https://github.com/mintapi/mintapi/pull/392, a constant was created for the soft token MFA Method.  However, an underscore was used (i.e. `soft_token`) instead of a dash (i.e. `soft-token`).


mrooney (approved):
Thanks!

PR #455, opened 2022-04-23
[FIX] - Add exception to wait in MFA Loop

burkematthew:
When using the WebDriver wait in the MFA Loop, the way it's currently written, we are assuming that the page WILL load.  However, because a user may have multiple MFA entries required, we may actually validly loop back to the top and the page will not load.  Without the try\except, MintAPI crashes.

### DIFF
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -368,10 +368,14 @@ def sign_in(
         )
         account_selection_page(driver, intuit_account)
         password_page(driver, password)
-        # Give the overview page a chance to actually load
-        WebDriverWait(driver, 5).until(
-            expected_conditions.url_contains("https://mint.intuit.com/overview")
-        )
+        # Give the overview page a chance to actually load.
+        # If it doesn't, then there may be another round of MFA.
+        try:
+            WebDriverWait(driver, 5).until(
+                expected_conditions.url_contains("https://mint.intuit.com/overview")
+            )
+        except Exception:
+            pass
 
     driver.implicitly_wait(20)  # seconds
     # Wait until the overview page has actually loaded, and if wait_for_sync==True, sync has completed.
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -337,6 +337,7 @@ def sign_in(
 
     user_selection_page(driver)
 
+    count = 0
     while not driver.current_url.startswith("https://mint.intuit.com/overview"):
         try:  # try to enter in credentials if username and password are on same page
             handle_same_page_username_password(driver, email, password)
@@ -375,7 +376,11 @@ def sign_in(
                 expected_conditions.url_contains("https://mint.intuit.com/overview")
             )
         except Exception:
-            pass
+            count += 1
+            if count > 10:
+                raise RuntimeError(
+                    "Login to Mint failed due to timeout in the Multifactor Method Loop"
+                )
 
     driver.implicitly_wait(20)  # seconds
     # Wait until the overview page has actually loaded, and if wait_for_sync==True, sync has completed.
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -377,7 +377,7 @@ def sign_in(
             )
         except Exception:
             count += 1
-            if count > 10:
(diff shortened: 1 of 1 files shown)


mrooney (commented):
Makes sense! My only concern is, does this introduce the potential for an infinite loop / hang now? If we end up in an unexpected scenario, would it loop forever, or would it hit an error and error out?

tsdg112:
I don't understand why you want the WebDriverWait.  The while statement at the top of the loop ensures that the loop will end when the overview page is loaded.  The WebDriverWait slows things down by adding an 5 second wait to iterate the loop.

@mrooney Take a look at PR #452.  It introduces a loop counter to ensure that the loop eventually terminates.

tsdg112:
I don't have a phone number associated with my Mint account, so I always need the loop to iterate at least once for bypass_verified_user_page to process.  Right now the WebDriverWait only waits until you are logged in.  Can you change it so that the wait also terminates when any of the conditions that require the loop to be repeated occur (such as the existence of ius-verified-user-update-btn-skip to bypass the "Let's add your current mobile number" page).

burkematthew:
@tsdg112 the reason for the wait is because I've experienced the script running faster than the browser has a chance to load.  I could potentially reduce the length of the wait, but the sole purpose is to address the race condition.

And to prevent the infinite loop, I can add a counter similar to what your PR does @tsdg112 .

burkematthew:
> I don't have a phone number associated with my Mint account, so I always need the loop to iterate at least once for bypass_verified_user_page to process. Right now the WebDriverWait only waits until you are logged in. Can you change it so that the wait also terminates when any of the conditions that require the loop to be repeated occur (such as the existence of ius-verified-user-update-btn-skip to bypass the "Let's add your current mobile number" page).

I don't mind that addition, though I'd like to keep that separate from this PR.  This PR is simply to address the fact that when there is a need to go back through the loop, it's actually crashing right now.

tsdg112:
If you successfully log in in the first iteration of the loop, it will crash when handle_different_page_username_password throws an exception in the second iteration.  This needs to be fixed, too.

burkematthew:
@tsdg112 are you saying that does happen for you on this branch?  I've tried with one MFA method (i.e. email) and multiple MFA methods (i.e. email and SMS) and both successfully finish.  

I'll add in the counter to ensure that this doesn't hang, but other than that, I think this PR should be G2G.

tsdg112:
@burkematthew Yes, this happened to me when I was debugging PR 452.  It happens because Mint always directs me to the "Let's add your current mobile number" page, which is bypassed by bypass_verified_user_page in the next loop iteration.  Did you test with a login that elicits this interaction with Mint?

burkematthew:
@tsdg112 I haven't come across that yet, but I'll work with my Mint account and see if I can get it to appear.

That being said, these changes solve a different problem.  I know your PR that you opened attempted to solve the problem you're primarily concerned about, so I think we can address the scenario you're referring to there.

tsdg112:
@burkematthew OK, thats fine with me.

mintapi/signIn.py
```diff
@@ -368,10 +369,18 @@ def sign_in(
...
-        )
+        # Give the overview page a chance to actually load.
+        # If it doesn't, then there may be another round of MFA.
+        try:
+            WebDriverWait(driver, 5).until(
+                expected_conditions.url_contains("https://mint.intuit.com/overview")
+            )
+        except Exception:
+            count += 1
+            if count > 10:
```
mrooney, line 380: Nice! Do we actually want to hold this up for 10 loops (almost a minute)? This was implemented to handle a 2nd loop IIRC, so maybe 3 would be sufficient? Or 4-5?

mrooney (changes requested):

mintapi/signIn.py
burkematthew: Yeah fair point! I think 4 is good for now.

burkematthew: @mrooney done!

mrooney (approved):
Awesome job, thanks!

PR #460, opened 2022-04-27
fix get_net_worth()

dzg:


burkematthew:
@dzg what kind of account shows up without a current balance? Just curious for better understanding of why this is needed.

burkematthew:
@dzg what kind of account shows up without a current balance? Just curious for better understanding of why this is needed.

burkematthew (approved):

Issue #458, opened 2022-04-26
Mint Initialization not working


gerber55555:
**Describe the bug**
Mint Initialization hangs after login, non-headless version shows login successful, but it never moves on to next step.

What version of MintAPI are you using?
1.66

What command[s] did you run / steps to reproduce?
on cmdline:
`mintapi --keyring <email> --use-chromedriver-on-path `

in python
```
mintapi.Mint(...) <-- Hangs here
mint.get_accounts()
```

Stacktrace/error received:
None

**What did you expect to happen?**
Expectation of mint api to return account information

**What actually happened?**
No data returned, hangs indefinitely with no error 

**Additional Context**
Ran on RaspberryPi and Windows 10. Using Python 3.9.2

gerber55555:
Apparently there is a new version of Mint that they are working on to support. If you pull the repo and build it locally you will be able to use it. Keeping this issue open until new version is released however.

nickodell:
@gerber55555 Do you have this working? If so, which branch do you have working on? I tried branch main, at commit bc6e78c, and I ran into this error:

```
Traceback (most recent call last):
  File "./fetch.py", line 110, in <module>
    main()
  File "./fetch.py", line 60, in main
    accounts = mint.get_accounts()
AttributeError: 'Mint' object has no attribute 'get_accounts'
```

I'm not exactly sure why this method was removed - it's still mentioned in the README. It looks like it was removed as part of a refactor in PR #430.

Is get_account_data() the replacement for the old get_accounts()?

gerber55555:
@nickodell Yes I got it working, I pulled the same branch and commit as you. And it seems like get_account_data() is the replacement for get_accounts(), which provided the data I needed. I'm not sure why it was changed either.

mrooney:
@burkematthew do you think it would make sense to write a wrapper method so folks can still use `get_accounts()`? Backwards imcompatible changes seem fine if super helpful, but this might confuse a bunch of folks? Or maybe we just need to document it in the CHANGELOG.

burkematthew:
@mrooney I missed making the changes in the README, which is part of the confusion.

One of the reasons I felt it was helpful to break the names is that each method has different input and very likely different outputs. By "breaking" the old method, we force users to take a look at the new input and the structure of the new output instead of them going through trial and error on what changed to make it break. 

That being said, if the consensus is wrappers would still be beneficial, then that's something we can do.

nickodell:
[...]

PR #462, opened 2022-05-01
[CHORE] - Update the README, CHANGELOG and more tidying of method names

burkematthew:
As discussed on #458 , the names of the methods have changed from v1 and so we need some updates to the README, more explicit callout in the CHANGELOG, and we need to make some more tweaks\tidying to the actual method names to ensure consistency.  Note that `get_bills` is an exception because it does NOT use the same execution structure as the others.


mrooney:
Thanks! Am I correct in understanding that all the methods which you've renamed in this PR have had their input and/or output changed and they are being renamed to be consistent with the other breaking method changes?

burkematthew:
Yes and no. Some of them don't have any output changes, but I thought it was important to keep consistency in the naming convention across our methods. Improves usability so that one method isn't `get_budgets` and another is `get_account_data`. The remaining ones I changed in this PR was more about that consistency.

burkematthew:
@mrooney any other comments?

mrooney (approved):
Thanks, looks good! Before we release, I'll do a re-review of the changes and adjust some things in prominence; this looks great for now!

PR #470, opened 2022-05-13
[FEATURE] - Add support for routing to Mint beta

burkematthew:
The biggest pain point of Mint's rollout of the new UI is not everyone has the rollout.  Thanks to a user's observation, the URL of `beta.mint.intuit.com` does route to the new UI.  So, this PR adds support for the tool to route to the beta URL instead of the normal URL, with the hopes of giving support for those users whose accounts are still on the old UI.

### DIFF
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -317,14 +317,20 @@ def sign_in(
     imap_password=None,
     imap_server=None,
     imap_folder="INBOX",
+    beta=False,
 ):
+    if beta:
+        url = constants.MINT_BETA_ROOT_URL
+    else:
+        url = constants.MINT_ROOT_URL
     """
     Takes in a web driver and gets it through the Mint sign in process
     """
     driver.implicitly_wait(20)  # seconds
-    driver.get("https://www.mint.com")
-    element = driver.find_element_by_link_text("Sign in")
-    element.click()
+    driver.get(url)
+    if not beta:
+        element = driver.find_element_by_link_text("Sign in")
+        element.click()
 
     WebDriverWait(driver, 20).until(
         expected_conditions.presence_of_element_located(
@@ -340,7 +346,7 @@ def sign_in(
 
     driver.implicitly_wait(1)  # seconds
     count = 0
-    while not driver.current_url.startswith("https://mint.intuit.com/overview"):
+    while not driver.current_url.startswith("{}/".format(url)):
         try:  # try to enter in credentials if username and password are on same page
             handle_same_page_username_password(driver, email, password)
         except (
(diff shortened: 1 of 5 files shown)


mintapi/signIn.py
```diff
@@ -317,14 +317,20 @@ def sign_in(
...
+        url = constants.MINT_ROOT_URL
     """
     Takes in a web driver and gets it through the Mint sign in process
     """
     driver.implicitly_wait(20)  # seconds
-    driver.get("https://www.mint.com")
-    element = driver.find_element_by_link_text("Sign in")
-    element.click()
+    driver.get(url)
+    if not beta:
```
burkematthew, line 331: Referencing the beta URL takes you right to the sign-in page, so this is not needed in that instance.

jiehanzheng:
I tested out your branch, and it is working perfectly for me (besides having to adapt to the new format).

I've verified that these features work:
- `mintapi.Mint(..., beta=True)`
- `mint.get_transaction_data`
- `mint.get_account_data`

Thanks for working on this!

mrooney:
Nice! Since this would go in 2.0, you'd always want this if you have the old UI right? I'm wondering if it makes sense just to always use beta for now so there's no need for a switch? Then maybe a follow-up PR could detect the old UI and restart the flow with beta? Since we have the ability, it seems nice if the 2.0 branch just "works" instead of requiring documentation around a switch. What do you think?

jiehanzheng:
> if it makes sense just to always use beta for now so there's no need for a switch?

What do you think if we change the default to `beta=True` but still keep the non-beta code path there?  If Mint removes the domain redirect, sending everyone to beta will cause everyone's mintapi to break and they'd have to to downgrade mintapi or wait for the next version that fixes this.

Whereas if we keep the non-beta code path behind a flag, new UI users can opt themselves out of this by something like `--no-beta` / `beta=False` as a quick mitigation

burkematthew:
I tend to agree with @jiehanzheng . I wouldn't want us to just use the beta URL without an option to turn it off. I can flip the direction so that the default is true and if you don't want to use it, you can supply a flag.

burkematthew:
Actually, after thinking about this a little more, I think it's better the way it is.  Considering that v2 corresponds with compatibility specifically with the new UI, it makes more sense to have a flag that allows you to override your existing Mint settings and access the new UI.  This way, as a user's account migrates to the new UI, they can rely on their own runtime options.  Lastly, think about it this way: once everyone is on the new UI, it wouldn't make sense to have everyone remove this option from their implementation.  With this direction, it will naturally fall out of use.

@jiehanzheng @mrooney

mrooney (changes requested):
Seems reasonable enough! Do you mind just adding a sentence to the top part of the README about 2.x vs 1.x that mentions you can also use `--beta` if you have the old UI? I think that will help!

burkematthew:
@mrooney done!

Issue #467, opened 2022-05-11
Multi-Data Support? Huh?


dzg:
Can someone explain [this?](https://github.com/mintapi/mintapi#multi-data-support)
How do I get "multiple types of data in one call to mintapi"?

mrooney:
It means you can pass multiple arguments for the data you want, instead of just one, and it can write out multiple files. Perhaps we need an example call there :)

dzg:
Yes an example would be super 
Thanks!

mrooney:
@dzg let me know if #471 seems sufficient!

PR #471, opened 2022-05-15
explain multi-data with examples

mrooney:
Attempting to address #467 with some concrete arguments and examples. @burkematthew let me know if I've gotten anything wrong :)

I also changed `investments` to `investment` in the docs as I think that is the correct argument? But let me know if I've made a mistake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ While Mint supports authentication via Voice, `mintapi` does not currently suppo
 
 ### Multi-Data Support
 
-As of v2.0, mintapi supports returning multiple types of data in one call to mintapi.  When exporting multiple data types, you can either send it directly to `stdout` or you can export to a data file.  What mintapi will do with your specified filename is add a suffix based on the type of data you are exporting.  The following table outlines the option selected and its corresponding suffix:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,20 +109,19 @@ While Mint supports authentication via Voice, `mintapi` does not currently suppo
...
+As of v2.0, mintapi supports returning multiple types of data in one call, such as: `mintapi --accounts --budgets --transactions`.  When exporting multiple data types, you can either send it directly to `stdout` or you can export to a file via `--filename`.  mintapi will create a file for each type of data, with a suffix based on the format.  For example, if you run `mintapi --accounts --transactions --filename=current --format=csv`, then you will receive two files: `current_account.csv` and `current_transaction.csv`.  The following table outlines the option selected and its corresponding suffix:
 
 | Option       | Suffix       |
 | -----------  | -----------  |
 | accounts     | account      |
 | budgets      | budget       |
 | transactions | transaction  |
 | categories   | category     |
-| investments  | investment   |
+| investment   | investment   |
```
burkematthew, line 120: @mrooney this should correctly be `investments`.  See the CLI argument definition for source of truth.

burkematthew (changes requested):
`investments` was correct.

https://github.com/mintapi/mintapi/blob/2285c840d06ce03abaea1b849dce6e37e155ab24/mintapi/cli.py#L173

README.md
mrooney: @burkematthew thanks! I tweaked the usage instead, let me know if that looks right :)

burkematthew (approved):

PR #472, opened 2022-05-15
[FIX] - Format Credit Score Data

burkematthew:
Similar to #460 , we need to format the data for the credit score so that when we export to a CSV, the function operates correctly.


mrooney (approved):
Looks great, thanks!

PR #476, opened 2022-05-29
Support optional MFA if remember my device is enabled

zsolteszku:
If you enable "Remember my device" then MFA will become optional. If you specify the preferred MFA method through the `mfa-method` setting then if MFA page won't appear then you'll get a `RuntimeException`. When "Remember my device" is enabled MFA page is not expected to show up.

### DIFF
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -510,7 +510,12 @@ def mfa_page(
     if mfa_method is None:
         mfa_result = search_mfa_method(driver)
     else:
-        mfa_result = set_mfa_method(driver, mfa_method)
+        try:
+            mfa_result = set_mfa_method(driver, mfa_method)
+        except Exception as e:
+            # MFA is optional for devices that were registered to Mint by clicking on "Remember my device"
+            logger.info(str(e))
+            return
     mfa_token_input = mfa_result[0]
     mfa_token_button = mfa_result[1]
     mfa_method = mfa_result[2]


mintapi/signIn.py
```diff
@@ -510,7 +510,12 @@ def mfa_page(
     if mfa_method is None:
         mfa_result = search_mfa_method(driver)
     else:
-        mfa_result = set_mfa_method(driver, mfa_method)
+        try:
+            mfa_result = set_mfa_method(driver, mfa_method)
+        except Exception as e:
```
mrooney, line 515: Is there a way we can actually detect the success, or catch a more specific exception, vs just catching all exceptions?

zsolteszku: The `set_mfa_method` catches the actual exception and re-raising a `RuntimeException`. So I feel we have three options:
1. Create a specific exception type for this that should be thrown from the `set_mfa_method` and caught here. I personally prefer this option.
2. Instead of re-raise a `RuntimeException` in the `set_mfa_method` we could just re-raise the actual exception so we can keep the actual type. The problem with this would be that in this part of the code that type of the exception would make not that much sense (`NoSuchElementException`).
3. I can catch `RuntimeException` here

Any thoughts on which one would you prefer or any other ideas for other options?

burkematthew: I vote option 1.

burkematthew:
While this is mainly about what happens if a user has already selected to remember the device, could you also add handling for when a user gets that screen in the first place? So this PR could be the entire handling of the "remember my device" feature of Mint's login.

zsolteszku:
> While this is mainly about what happens if a user has already selected to remember the device, could you also add handling for when a user gets that screen in the first place? So this PR could be the entire handling of the "remember my device" feature of Mint's login.

The way how we could handle it is by clicking the "Remember me" checkbox on login. In my personal opinion this does not need to be supported through mintapi, as this is a one time setting that the user can click on their device once they log in. Handling it programatically is not really beneficial (you do not need to click that checkbox at every login). It would just add complexity to the CLI (as I assume this would be a setting) without significant benefits.

What is important that if the user already clicked it, we still can fetch the data. The current version of mintapi unfortunately crashes if a user ever clicked that checkbox and not sure that you can "uncheck it". I recently updated to mintapi v2.0, in v1.0 looks like that MFA was not required (did not raise exception if the MFA method was not found).

mrooney:
That's fair, I'm good with this approach if you are @burkematthew !

mrooney (approved):

burkematthew:
@zsolteszku @mrooney I'm not sure I agree. If the user encounters this screen, MintAPI would either error out with some unrelated error or it would hang leaving the user wondering what's up. The latter is the worst of the two because it would require them to not use headless mode and troubleshoot to see what's happening. This to me is a poor UX and puts too much on the casual user that may not even know to do that. 

My thought is that we should, by default, remember the device. So we need handling to check for this screen and if we find it, select to remember the device. We should also include an option that allows users to select to NOT remember the device, similar to how we have an option to not wait for the sync to complete. This then becomes minimally obtrusive to the typical user while still allowing us to properly handle the screen.

burkematthew:
@zsolteszku my apologies, I understand now. There was a similar screen that allows some bypassing and I was confusing what you were referring to with that. 

I'm good now!

PR #484, opened 2022-06-10
[FIX] - remove duplicate MFA method

hudcap:
I'm running into an issue when supplying the `mfa_token`
I traced the issue to there being an `mfa_method` called "soft-token" and another called "authenticator"
`soft-token` is missing its `span_css_selector` (empty string) and also the word "soft-token" does not appear in the span text, so
```python
result = mfa_method in span_text
```
yields `False`

My solution was to just get rid of the soft-token method and leave the authenticator method.
This might break backwards-compatibility.
Not sure which name you want to keep.

Sorry for the multiple commits that got attached—still haven't quite figured out this rebasing business. I assume they all get squashed for the PR, but please let me know if not.


burkematthew:
@hudcap thanks for the contribution! Glad to discuss it here. 

So, up until the dynamic MFA was put in place, there wasn't MintAPI support for using an Authenticator app. What it did support was a token based authentication using a runtime-supplied token, which was the soft -token. 

I see your description talks about what you are observing but it doesn't talk much about your true use case. Am I correct that you ARE using an Authenticator App to supply an MFA code and this is what is not working? If that's the case, then I think the PR should be revisited because we really hadn't put in full support for using an Authenticator app yet. We included some bare bones to accompany other MFA methods but it hadn't been fully fleshed out. 

If that's not the case, then I'd love to get some more details about your use case. 

Again, thanks for your continued; we appreciate it!!

hudcap:
My pleasure—thank _you_ for all the work you put into this!

Starting with my use case, since I'm not sure I understand the difference between soft-token and authenticator app.
I use the Authy app, which does an initial sync-up and then generates a new 6-digit code every 30 seconds. I pass the initial sync-token (whatever that's called) to `mfa-token` and `mintapi` generates that 6-digit code. Until now, I've been passing "soft-token" to the `mfa-method` parameter, but in the new version that argument isn't supposed to be passed if there's only one MFA method.
I started getting errors and discovered 2 issues:

1. This is an empty string and so it doesn't match anything:
https://github.com/mintapi/mintapi/blob/178db0a94af0ce114f1a9f473c614227aeb37d38/mintapi/signIn.py#L48
2. This check:
https://github.com/mintapi/mintapi/blob/178db0a94af0ce114f1a9f473c614227aeb37d38/mintapi/signIn.py#L566
uses this value:
https://github.com/mintapi/mintapi/blob/178db0a94af0ce114f1a9f473c614227aeb37d38/mintapi/constants.py#L19
but the `span_text` only contains the word "autheticator" and not "soft-token"

My PR was predicated on the assumption that there is only one such MFA method, but it seems that is incorrect.

mintapi/signIn.py
```diff
@@ -41,13 +41,6 @@ class MFAMethodNotAvailableError(RuntimeError):
 BUTTON_CSS_SELECTORS_LABEL = "button_css_selectors"
 
 MFA_METHODS = [
-    {
```
burkematthew, line 44: This needs to stay.

mintapi/signIn.py
```diff
@@ -530,7 +523,7 @@ def mfa_page(
         return
 
     # mfa screen
-    if mfa_method == constants.MFA_VIA_SOFT_TOKEN:
+    if mfa_method == constants.MFA_VIA_AUTHENTICATOR:
```
burkematthew, line 526: This shouldn't change.  This should correctly be for soft token.

burkematthew:
@hudcap The use of `soft-token` is for when you build up a token and pass it into the Mint object.  This is different than if you use something like Google Authenticator, which provides an OTP, just like you may get in text or email.

The use of the authenticator app should be captured by `handle_other_mfa`.  Can you provide more information on what errors you are getting?

burkematthew (changes requested):

burkematthew:
@hudcap just wanting to ping you so you knew this is awaiting changes from you.

hudcap:
Sorry about that—this is sitting in my inbox waiting for me to do.
Been very busy lately and to test this out, I need to set up a clean copy with all my info—haven't gotten to it yet, but hope to soon.
Thanks!

To clarify, what do you mean by "build up a token?"
I pass a token to `mfa-token`, which is the same token my Authy app has that generates OTPs every 30 seconds. When I looked at the `mintapi` source code, it appeared to be doing the same thing and generating the same OTP as Authy.
soft-token and authenticator app still seem the same to me—where am I going wrong?

burkematthew:
@hudcap Really I was trying to make a distinction between supplying the token at invocation and supplying the token when prompted by Mint. I'm wondering if it's not an either/or put a potential AND. Meaning, we can look for soft token or authenticator and apply the token provided at invocation.

mintapi/signIn.py
hudcap: I can leave soft-token alone and fix the selector and also change the method name so it matches if you are more comfortable with that. But I think that would make it a duplicate of the Authenticator dict...

hudcap:
> @hudcap Really I was trying to make a distinction between supplying the token at invocation and supplying the token when prompted by Mint. I'm wondering if it's not an either/or put a potential AND. Meaning, we can look for soft token or authenticator and apply the token provided at invocation.

I'm very much inclined to agree with this.
From the mint perspective, I believe they are the same. The user can handle this method in 2 different ways.

mintapi/signIn.py
hudcap: @burkematthew what are your thoughts on this?

hudcap:
@mrooney @burkematthew Are either of you able to review this? I just tried the latest version and am getting the same errors without my fixes.
Is mfa_token working for anyone as-is?

hudcap:
Looking at the new documentation from #583, it seems that setting `mfa_method` to `soft-token` is required for this to work, despite other documentation stating that it's only necessary if there are multiple authentication method options. It now works, so I will close this.
Thanks and sorry!

mrooney:
Thanks for the update! Perhaps a PR to the README/docs would be appropriate then?

hudcap:
@mrooney I was going to, but I'm not confident about whether/when `mfa_method` is actually required. Since I only use `soft-token`, I don't have any other data points.

PR #487, opened 2022-06-18
[CHORE] - Update README with content about the email nuance

burkematthew:
See the linked Issue for motivation of the clarification.


mrooney (changes requested):
Seems good! Super nitpick but could you change it from "I could" to "you could"? I don't think we use "I" anywhere else in the README but let me know if I'm mistaken.

burkematthew:
Yup you're right! I'll get it fixed.

burkematthew:
@mrooney done!  And found one other instance that I corrected as well.

PR #488, opened 2022-06-18
[FEATURE] - Add an option to raise an exception if the data refresh fails

burkematthew:
As requested by @felciano , we have added handling so that users can request to exit prematurely if the data sync times out.  This PR includes an implementation for the Command Line as well.


README.md
```diff
@@ -268,6 +269,7 @@ Run it as a sub-process from your favorite language; `pip install mintapi` creat
       --limit               Number of records to include from the API.  Default is 5000.
       --show-pending        Retrieve pending transactions.
                             Used with --transactions
+      --fail-if-stale       At login, Mint attempts to refresh your data.  If you wish to exit when the fail syncs, use this option.
```
mrooney, line 272: Should this say "when the sync fails"? Or, am I just reading it poorly?

mrooney (changes requested):
Nice approach, looks great! Just one QQ.

README.md
burkematthew: Yup, I thought I caught all of those instances.

mrooney (approved):
LGTM! I'm curious if the original request was to have a specific exit code that would be unique to this error, but if so, that's very easy to fix in a follow-up :)

PR #489, opened 2022-06-18
[CHORE] - Intentionally switch from mfa_method to mfa_type

burkematthew:
When we moved to the dynamic flow for Multifactor Authentication at login, many users reported issues because they were still sending mfa_method even though most users should not need this as MintAPI detects which MFA Type is requested\needed.  This change, though breaking in a minor version, is intended to force users into reading the docs and only sending this if needed.  Hopefully this reduces the number of issues reported on the tool hanging as a result of incorrectly using the argument.


mrooney:
Interesting idea! There are some folks who are intentionally using this right, and I've seen some reports saying that using this fixes an issue for them? So I'm hesitant to break things again for them. Could we solve the same thing by displaying a prominent warning when using it that isn't recommended and may cause issues? This warning could then be removed in a few versions.

burkematthew:
I'm going to close for the time being.

PR #490, opened 2022-06-19
[CHORE] - Add WebDriverWait for Account Selection Selectors

burkematthew:
Users reported that MintAPI often hangs on the Account Selection Page.  Mint's flow loads too quickly and so the continue button doesn't have a chance to load before we attempt to find and click it.  Add a `WebDriverWait` so that we can make sure all items have loaded first.


mintapi/signIn.py
```diff
@@ -658,7 +658,14 @@ def account_selection_page(driver, intuit_account):
                 )
             )
             account_input.click()
-
+        WebDriverWait(driver, 2).until(
```
mrooney, line 661: Seems like a good improvement! Two questions:

1. What is the reasoning for 2 seconds? Was that sufficient for folks who ran into the error? It could be longer but I'm assuming we don't want to delay everyone who may not be seeing this page?
2. Will the exception catching still work as desired below? I see a TimeException [can be thrown](https://selenium-python.readthedocs.io/waits.html) so I wasn't sure if that needs to be updated.

burkematthew: @mrooney 2 seconds was sufficient for my observations and figured we could always increase that if need be. 

Yes the NoSuchElement is still good because we wait until it's loaded and if it's not, the element doesn't exist.

mrooney: > @mrooney 2 seconds was sufficient for my observations and figured we could always increase that if need be.
> 
> Yes the NoSuchElement is still good because we wait until it's loaded and if it's not, the element doesn't exist.

Got it, I just wanted to make sure that wasn't going to throw a `TimeoutException`, but it sounds like I'm reading that wrong: `If no element is found in that time, a TimeoutException is thrown`. If it will happily continue on, that sounds good! If not, I we'd just need to add TimeoutException along with the NoSuchElement one. Approving assuming I understand you correctly!

burkematthew: Nope, we should be good!

Issue #492, opened 2022-06-23
The sign in page doesn't load fast enough and the click to sign in fails


gabefortin:
**Describe the bug**
Not logging in to Mint. Get error 'selenium.common.exceptions.WebDriverException: Message: unknown error: cannot determine loading status' and Chrome shuts down instantly.

What version of MintAPI are you using?
2.1

What command[s] did you run / steps to reproduce?
Stacktrace/error received:
Message: unknown error: cannot determine loading status
from unknown error: unexpected command response
  (Session info: chrome=103.0.5060.53)
Traceback (most recent call last):
  File "/Users/gabe.fortin/Documents/coding/budgettracker/venv/lib/python3.9/site-packages/mintapi/api.py", line 166, in login_and_get_token
    self.status_message = sign_in(
  File "/Users/gabe.fortin/Documents/coding/budgettracker/venv/lib/python3.9/site-packages/mintapi/signIn.py", line 333, in sign_in
    element.click()
  File "/Users/gabe.fortin/Documents/coding/budgettracker/venv/lib/python3.9/site-packages/selenium/webdriver/remote/webelement.py", line 80, in click
    self._execute(Command.CLICK_ELEMENT)
  File "/Users/gabe.fortin/Documents/coding/budgettracker/venv/lib/python3.9/site-packages/selenium/webdriver/remote/webelement.py", line 633, in _execute
    return self._parent.execute(command, params)
  File "/Users/gabe.fortin/Documents/coding/budgettracker/venv/lib/python3.9/site-packages/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/Users/gabe.fortin/Documents/coding/budgettracker/venv/lib/python3.9/site-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.WebDriverException: Message: unknown error: cannot determine loading status
from unknown error: unexpected command response
  (Session info: chrome=103.0.5060.53)

Traceback (most recent call last):
  File "/Users/gabe.fortin/Documents/coding/budgettracker/venv/lib/python3.9/site-packages/mintapi/api.py", line 166, in login_and_get_token
    self.status_message = sign_in(
  File "/Users/gabe.fortin/Documents/coding/budgettracker/venv/lib/python3.9/site-packages/mintapi/signIn.py", line 333, in sign_in
    element.click()
[...]

PR #493, opened 2022-06-24
[FIX] - Add a delay before the clicking sign in button

CalvinChanCan:
This is a fix for the issue #492 where selenium generates an error when clicking the sign in button.

### DIFF
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -335,11 +335,6 @@ def sign_in(
     driver.implicitly_wait(20)  # seconds
     driver.get(url)
     if not beta:
-        WebDriverWait(driver, 20).until(
-            expected_conditions.presence_of_element_located(
-                (By.LINK_TEXT, "Sign in")
-            )
-        )
         # Add 1 second delay otherwise an issue occurs when trying to click the sign in button on home page
         time.sleep(1)
         home_page(driver)
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -335,8 +335,11 @@ def sign_in(
     driver.implicitly_wait(20)  # seconds
     driver.get(url)
     if not beta:
-        element = driver.find_element_by_link_text("Sign in")
-        time.sleep(1)
+        element = WebDriverWait(driver, 20).until(
+            expected_conditions.presence_of_element_located(
+                (By.LINK_TEXT, "Sign in")
+            )
+        )
         element.click()
 
     WebDriverWait(driver, 20).until(
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -340,6 +340,8 @@ def sign_in(
                 (By.LINK_TEXT, "Sign in")
             )
         )
+        # Add 1 second delay otherwise an issue occurs when trying to click the sign in button on home page
+        time.sleep(1)
         home_page(driver)
         
     WebDriverWait(driver, 20).until(
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -335,11 +335,12 @@ def sign_in(
     driver.implicitly_wait(20)  # seconds
     driver.get(url)
     if not beta:
-        element = WebDriverWait(driver, 20).until(
+        WebDriverWait(driver, 20).until(
             expected_conditions.presence_of_element_located(
                 (By.LINK_TEXT, "Sign in")
             )
         )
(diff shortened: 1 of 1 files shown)


mintapi/signIn.py
```diff
@@ -336,6 +336,7 @@ def sign_in(
     driver.get(url)
     if not beta:
         element = driver.find_element_by_link_text("Sign in")
+        time.sleep(1)
```
burkematthew, line 339: Instead of sleeping, you should use WebDriverWait for the element in question. That way there isn't a 1 second sleep unless it's required.

burkematthew (changes requested):

mintapi/signIn.py
CalvinChanCan: I do agree that its better to avoid using `time.sleep` when possible. However, when trying to using WebDriverWait, the error still occurs. So its odd that its finding the sign in button but the click is not working without a delay. Its also a bit difficult to debug since a linebreak before the click gives it enough time to work as expected.

```
        element = WebDriverWait(driver, 20).until(
            expected_conditions.presence_of_element_located(
                (By.LINK_TEXT, "Sign in")
            )
        )
```

burkematthew: @CalvinChanCan can you commit the version with WebDriverWait that doesn't solve it? Just want to take a look and see if anything sticks out as to why.

CalvinChanCan: Thanks @burkematthew! I made committed the version with WebDriverWait that doesn't resolve.

mintapi/signIn.py
```diff
@@ -335,7 +335,11 @@ def sign_in(
     driver.implicitly_wait(20)  # seconds
     driver.get(url)
     if not beta:
-        element = driver.find_element_by_link_text("Sign in")
+        element = WebDriverWait(driver, 20).until(
```
burkematthew, line 338: @CalvinChanCan If I'm not mistaken, WebDriverWait doesn't actually return the element but I think a bool value. I'd put the wait above the original find element and see if that works.

CalvinChanCan: @burkematthew From the Selenium [docs](https://www.selenium.dev/documentation/webdriver/waits/#:~:text=el%20%3D%20WebDriverWait(driver%2C%20timeout%3D3).until(lambda%20d%3A%20d.find_element_by_tag_name(%22p%22))), WebDriverWait can return the element if it is found and be interpreted as a truthy bool value. Though I have tried as suggested and still its still erroring on the click but does proceed to the sign in page though 😕.

burkematthew: @CalvinChanCan interestingly enough (though I couldn't really explain why), if you move the assignment of element and the click to its own method, say, home_page, and put the WebDriverWait before you call home_page, it seems to work as expected:

```
WebDriverWait...
home_page(driver)
```

CalvinChanCan: @burkematthew Do you mean how I have it in the latest commit? For me, I'm still experiencing the same error still.

burkematthew: @CalvinChanCan that's super weird as it resolved the issue for me. 

I guess the time.sleep will have to suffice then. Thanks for trying out the other options!

By the way, go ahead and just leave the home_page method in place. A little cleaner anyways!

CalvinChanCan: @burkematthew, yeah it is strange how its resolving for some users and not for others. Though I have observed that sometimes mint inconsistently routing me to the new sign in page UI or to the old sign in page UI. The sign in page UI is completely new to me and I have never seen it until when this issue started occurring so I wonder if its related. 

image
image

Thanks again! I have kept the home_page method and added back `time.sleep`.

burkematthew: @CalvinChanCan you can take out the WebDriverWait since we're doing the explicit sleep.

CalvinChanCan: @burkematthew Done!

burkematthew (approved):

burkematthew (changes requested):
@CalvinChanCan there are some Lint issues. Can you please correct them and then we should be good to go?

CalvinChanCan:
> @CalvinChanCan there are some Lint issues. Can you please correct them and then we should be good to go?

Sorry about that. Should be fixed now.

burkematthew (approved):

Issue #499, opened 2022-06-28
Several issues during mintapi exploitation (mintapi not returning any data).


Mitrij:
**Describe the bug**
During mintapi exploitation faced several errors and flaws. Tracebacks follows.
Also about timeouts in several code parts, for slow computers waite timeouts in code are too small.
Can you check it and fix? Thanks.

mintapi 2.2

```
Traceback (most recent call last):
  File "/home/scrp/mintapi/mintapi/mintapi/api.py", line 169, in login_and_get_token
    self.status_message = sign_in(
  File "/home/scrp/mintapi/mintapi/mintapi/signIn.py", line 365, in sign_in
    handle_different_page_username_password(driver, email)
  File "/home/scrp/mintapi/mintapi/mintapi/signIn.py", line 451, in handle_different_page_username_password
    email_input = find_element(
NameError: name 'find_element' is not defined
```
should be driver.find_element instead of just find_element ?

```
Traceback (most recent call last):
  File "/home/scrp/mintapi/mintapi/mintapi/api.py", line 169, in login_and_get_token
    self.status_message = sign_in(
  File "/home/scrp/mintapi/mintapi/mintapi/signIn.py", line 394, in sign_in
    account_selection_page(driver, intuit_account)
  File "/home/scrp/mintapi/mintapi/mintapi/signIn.py", line 670, in account_selection_page
    WebDriverWait(driver, 20).until(
  File "/usr/local/lib/python3.9/dist-packages/selenium/webdriver/support/wait.py", line 80, in until
    raise TimeoutException(message, screen, stacktrace)
selenium.common.exceptions.TimeoutException: Message: 
```
Looks like 2 seconds not enough for slow computers.

```
Traceback (most recent call last):
  File "/home/scrp/mintapi/mintapi/mintapi/api.py", line 169, in login_and_get_token
    self.status_message = sign_in(
  File "/home/scrp/mintapi/mintapi/mintapi/signIn.py", line 365, in sign_in
    handle_different_page_username_password(driver, email)
  File "/home/scrp/mintapi/mintapi/mintapi/signIn.py", line 467, in handle_different_page_username_password
    for username_element in username_elements:
TypeError: 'WebElement' object is not iterable
```

Also, when it finally logs in it does not getting any data (by mint.get_account_data()).

iamjli:
I'm encountering a similar issue on mintapi 2.3

```
>>> mint = mintapi.Mint(username, password)
name 'find_element' is not defined
[...]

PR #502, opened 2022-06-29
[FIX] - Add missing driver to find_element to fix issue with signing in

CalvinChanCan:
This PR fixes the issue described in #499 since the `find_element` method is missing the driver.


mrooney (approved):
Thanks!

PR #503, opened 2022-06-29
bump to 2.4 to fix missing driver variable

mrooney:
Includes #502


mrooney:
I pushed this out to help folks but would still appreciate the review in case I missed something silly.

burkematthew (approved):

PR #507, opened 2022-06-30
[FEATURE] - Added bills download to mintapi cli

BillJr99:
Added a constants key for mint bills, a cli option --bills to download bill data from mint, and a call to mint.get_bills if that option is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ As of v2.0, mintapi supports returning multiple types of data in one call, such 
 | Option       | Suffix       |
 | -----------  | -----------  |
 | accounts     | account      |
+| bills        | bills        |
 | budgets      | budget       |
 | transactions | transaction  |
 | categories   | category     |
--- a/mintapi/cli.py
+++ b/mintapi/cli.py
@@ -455,7 +455,7 @@ def main():
         
     if options.bills:
         data = mint.get_bills()
-        output_data(options, data, constants.ACCOUNT_KEY, attention_msg)        
+        output_data(options, data, constants.BILL_KEY, attention_msg)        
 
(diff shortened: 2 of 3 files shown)


mrooney:
Nice, great addition, thanks! Was this manually tested/verified?

BillJr99:
You bet, and thanks for maintaining this!  I have not tested it extensively beyond my own mint account, but the cli does return the correct array of bills for my account.

BillJr99:
Looks like the lint check had failed, possibly due to line endings or whitespace.  I didn't have the action enabled on my fork.  I re-formatted with black and re-committed to this pull request if you'd like to re-run the checks on this end.

mrooney:
Thanks! For some reason GH shows the whole cli.py file as being rewritten. Is it possible spacing was changed, or changed to tabs, or similar?

BillJr99:
Looks like black changed all the line endings from unix to windows, but somehow only on that file!  I reverted to fix the line endings, and re-committed.  The linter also broke up a few help message string constants, but I left those in case that's preferrable.

BillJr99:
Sure thing.  I agree - I prefer not to have extraneous edits in a commit.  On my local repository, the black action now fails this commit because it still wants to break up those strings, but I'd assume it would continue to pass here.

burkematthew (approved):
Even though bills is still a bit out-of-sync with the rest of the endpoints, this by itself is good.  Thanks @BillJr99 for your contribution!

mrooney:
Thanks @BillJr99 ! Perhaps we need to have a contributor guide that specifies the version/arguments for black so it all aligns. Sorry for the troubles!

PR #510, opened 2022-07-05
[FIX] - Increase WebDriverWait for Account Selection Page

burkematthew:
Many users have reported a Timeout Exception related to the Account Selection Page.  When we (I) added the wait, 2 seconds was good enough at the time, but wanted to get feedback from users if it was long enough to wait.  At this point, we can safely say it was NOT long enough.  There also seems to be a difference in time needed in headless vs. non-headless mode.  Therefore, we are increasing from 2 to 20.

Future Enhancement Idea: Move all wait timeout periods to a variable and let users control this so that if one user needs less wait time than another, both can accommodate.


mrooney:
Thanks! Would this slow down any users who would not expect to see this element? Or does it not have any negative impacts besides slowing down a potential failure?

burkematthew:
@mrooney no it is just the maximum time to wait. If we find it sooner, it will continue forward!

mrooney (approved):

PR #514, opened 2022-07-12
[CHORE] - Move general filters to its own class instance

burkematthew:
In preparation for supporting the Transactions Search Endpoint, we should do a little refactoring to keep the Transactions changes as minimal as possible.  This PR addresses creating `filters.py`, which multiple classes can then reference instead of duplicating work.

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -8,13 +8,16 @@ from dateutil.relativedelta import relativedelta
 from mintapi import constants
 from mintapi.signIn import _create_web_driver_at_mint_com, sign_in
 from mintapi.trends import (
-    CategoryMatchFilter,
-    DateFilter,
-    DescriptionMatchFilter,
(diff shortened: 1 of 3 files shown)


mintapi/api.py
```diff
@@ -8,15 +8,18 @@
 from mintapi import constants
 from mintapi.signIn import _create_web_driver_at_mint_com, sign_in
 from mintapi.trends import (
-    CategoryMatchFilter,
-    DateFilter,
-    DescriptionMatchFilter,
+    DateFilter as TrendDateFilter,
```
burkematthew, line 11: This is done because I'm not yet sure if the Transaction Date Filter will be the same or not.  This way we can reference DateFilter in multiple classes.

mrooney: Might be cleaner to just import trends as use `trends.DateFilter` but I'm also fine with this approach!

PR #516, opened 2022-07-13
[FIX] email as the only MFA randomly fails and other issues

tsdg112:
New attributes were added to SELECT_CSS_SELECTORS_LABEL and SPAN_CSS_SELECTORS_LABEL in response to a new OTP selection page.

Mint sometimes sends the OTP with a MIME-encoded email. I added MIME decoding.

Since updating to Chrome 103, the script occasionally throws a WebDriverException when signing in which is now safely ignored.  According to Google, this is a known bug and the only fix is to downgrade to 102.

New attributes were added within the password_page function to handle a new password entry page.

This PR should fix some complaints about the runtime error "Login to Mint failed due to timeout in the Multifactor Method Loop".

### DIFF
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -116,16 +116,17 @@ def get_email_code(imap_account, imap_password, imap_server, imap_folder, delete
 
             msg = email.message_from_bytes(data[0][1])
 
+            x = email.header.make_header(email.header.decode_header(msg["Subject"]))
+
+            subject = str(x)
+            logger.debug("DEBUG: SUBJECT:", subject)
+
             x = email.header.make_header(email.header.decode_header(msg["From"]))
             frm = str(x)
             logger.debug("DEBUG: FROM:", frm)
 
             if not re.search("do_not_reply@intuit.com", frm, re.IGNORECASE):
                 continue
-
-            x = email.header.make_header(email.header.decode_header(msg["Subject"]))
-            subject = str(x)
-            logger.debug("DEBUG: SUBJECT:", subject)
 
             if not re.search("Your Mint Account", subject, re.IGNORECASE):
                 continue
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -705,11 +705,11 @@ def password_page(driver, password):
     try:
         driver.find_element(
             By.CSS_SELECTOR,
-            '#iux-password-confirmation-password, #ius-sign-in-mfa-password-collection-current-password'
+            "#iux-password-confirmation-password, #ius-sign-in-mfa-password-collection-current-password",
         ).send_keys(password)
         driver.find_element(
             By.CSS_SELECTOR,
-            '#ius-sign-in-mfa-password-collection-continue-btn, [data-testid="passwordVerificationContinueButton"]'
+            '#ius-sign-in-mfa-password-collection-continue-btn, [data-testid="passwordVerificationContinueButton"]',
         ).submit()
     except (
         NoSuchElementException,
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -704,10 +704,12 @@ def password_page(driver, password):
     # password only sometimes after mfa
     try:
         driver.find_element(
-            By.CSS_SELECTOR, '#iux-password-confirmation-password, #ius-sign-in-mfa-password-collection-current-password'
+            By.CSS_SELECTOR,
+            '#iux-password-confirmation-password, #ius-sign-in-mfa-password-collection-current-password'
         ).send_keys(password)
         driver.find_element(
-            By.CSS_SELECTOR, '#ius-sign-in-mfa-password-collection-continue-btn, [data-testid="passwordVerificationContinueButton"]'
+            By.CSS_SELECTOR,
+            '#ius-sign-in-mfa-password-collection-continue-btn, [data-testid="passwordVerificationContinueButton"]'
         ).submit()
     except (
         NoSuchElementException,
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -116,6 +116,13 @@ def get_email_code(imap_account, imap_password, imap_server, imap_folder, delete
 
             msg = email.message_from_bytes(data[0][1])
 
+            x = email.header.make_header(email.header.decode_header(msg["From"]))
+            frm = str(x)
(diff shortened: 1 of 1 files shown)


mintapi/signIn.py
```diff
@@ -118,13 +120,6 @@ def get_email_code(imap_account, imap_password, imap_server, imap_folder, delete
             subject = str(x)
             logger.debug("DEBUG: SUBJECT:", subject)
 
-            x = email.header.make_header(email.header.decode_header(msg["From"]))
-            frm = str(x)
-            logger.debug("DEBUG: FROM:", frm)
-
-            if not re.search("do_not_reply@intuit.com", frm, re.IGNORECASE):
```
burkematthew, line 125: @tsdg112 is there some sort of a domain we could check on the email?

mintapi/signIn.py
```diff
@@ -582,6 +582,23 @@ def search_mfa_method(driver):
...
+def check_mfa_method(driver):
+    for method in MFA_METHODS:
+        try:
+            mfa_token_select = driver.find_element(
+                By.CSS_SELECTOR, method[SELECT_CSS_SELECTORS_LABEL]
+            )
+        except (NoSuchElementException, ElementNotInteractableException):
+            continue
+        mfa_token_select.click()
+        driver.implicitly_wait(20)  # seconds
```
burkematthew, line 594: Can we do WebDriverWait instead?

mintapi/signIn.py
```diff
@@ -417,7 +414,10 @@ def sign_in(
 
 
 def home_page(driver):
-    element = driver.find_element(By.LINK_TEXT, "Sign in").click()
+    try:
+        element = driver.find_element(By.LINK_TEXT, "Sign in").click()
+    except WebDriverException:
+        logger.info("WebDriverException when clicking Sign In")
```
burkematthew, line 420: If we get this exception, are we safe to continue? Or should we exit the run?

mintapi/signIn.py
```diff
@@ -582,6 +582,23 @@ def search_mfa_method(driver):
     return mfa_token_input, mfa_token_button, mfa_method
 
 
+def check_mfa_method(driver):
```
burkematthew, line 585: I'm not quite sure I understand what this method addresses in the flow? Do you have any screenshots you could add to the PR?

mintapi/signIn.py
tsdg112: I think that we already have sufficient email validation after removing the from check.  If you still want to retain this check, the from field needs to be a configurable option, since it differs for each duck.com email address.

mintapi/signIn.py
tsdg112: This is what I get.  I only have one MFA method, so that I can't specify email as the MFA method.

[Intuit Accounts - Sign In.pdf](https://github.com/mintapi/mintapi/files/9104532/Intuit.Accounts.-.Sign.In.pdf)

mintapi/signIn.py
tsdg112: Yes, it is safe to ignore and continue.

mintapi/signIn.py
burkematthew: @tsdg112 I think this should be a separate PR isolated to a configurable option with the default as the email now.

tsdg112: Thats fine with me.  Should I restore the deleted code for this PR?
\

mintapi/signIn.py
burkematthew: @tsdg112 this is just the mfa_selection page, so I'd expect you to just pass mfa_method of email so that you get to the mfa_selection_page. If that's not working correctly, then we should fix that method instead. Otherwise, there is blind acceptance of the first one listed, but that not actually be the case.

mintapi/signIn.py
burkematthew: If we're waiting for the input CSS Selector, then we can do the wait here.

mintapi/signIn.py
burkematthew: @tsdg112 yes please! Thanks!

mintapi/signIn.py
tsdg112: The trouble is that Mint does not always present this screen.  I tried specifying the mfa_method, but when the screen is not presented, mintapi throws MFAMethodNotAvailableError.

burkematthew: @tsdg112 is there a way to first check if this screen exists by looking for the header text? Then, if it's there, we validate that mfa_method was supplied. If it's not there, then we proceed as usual.

tsdg112:
I could not reproduce the problem that required the new function check_mfa_method, so I removed the function.

mintapi/signIn.py
burkematthew: @tsdg112 can you restore the subject back to its original spot so that it doesn't show up as a diff in the PR?

burkematthew:
@tsdg112 can you also update the PR summary to reflect the current state of the changes?

tsdg112:
I updated the summary and reordered the statements.

mintapi/signIn.py
```diff
@@ -115,6 +117,7 @@ def get_email_code(imap_account, imap_password, imap_server, imap_folder, delete
             msg = email.message_from_bytes(data[0][1])
 
             x = email.header.make_header(email.header.decode_header(msg["Subject"]))
+
```
burkematthew, line 120: Can you eliminate this extra line?

mintapi/signIn.py
```diff
@@ -12,13 +12,15 @@
 import sys
 import time
 import zipfile
+import itertools
```
burkematthew, line 15: Is this still needed? I don't see any reference to it.

burkematthew (changes requested):

mintapi/signIn.py
tsdg112: Its for the next at line 149.

mintapi/signIn.py
tsdg112: Ok.  I removed it.

mintapi/signIn.py
burkematthew: Got it, thanks!

burkematthew (approved):

PR #523, opened 2022-07-15
[FEATURE] Show reason for login failures

tsdg112:
Detects login failures due to bad email, bad password, and CAPTCHA test and raises a RuntimeError with an appropriate message.


burkematthew (approved):

mrooney:
Great contribution, thanks! I imagine these will need to be updated from time to time to remain accurate, but it seems like a good step!

PR #530, opened 2022-07-21
[FIX] - Intuit Account Selection

burkematthew:
A couple things this PR addresses
* Unable to proceed past the account selection page.
* When an Intuit Account is supplied, we are not actually finding the radio button to select.

Bonus:
* CLI Support for supplying Intuit Account


mrooney (approved):
Awesome, thanks!

PR #531, opened 2022-07-24
[FEATURE] - Add support for accessing Transactions/Search API Endpoint

burkematthew:
So far in v2, we have had minimal transaction filtering because of the more complex API Endpoint layout.  However, thanks to the work of @eyadgaran, a lot of the legwork was done on how to best implement Filtering in API Requests.

This PR switches from the `Transactions` endpoint to the `Transactions/Search` endpoint and includes different types of filtering of transactions, such as:

* By Account ID
* By Category
* By Tag
* By Description
* By Date

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,9 +167,28 @@ As mentioned above, the Report Type is the type of report for which to generate 
 | 15         | Net Worth Over Time |
 | 16         | Net Income Over Time |
 
-#### Date Filter
+### Financial Data Transactions
 
-As mentioned above, the Date Filter is the date window for which to generate your trend analysis.  The supplied value must be one of the following enum values:
+If you want to provide a more granular filtering of your financial data transactions, you can select from a variety of search filters that are sent to Mint. 
+
+| Parameter         | Data Type          | Description  |
+| ----------------  | ------------------ | -----------  |
+| date_filter       | DateFilter.Options | The date window for which to filter your transactions. |
+| start_date        | Optional[str]      | An optional beginning date to your transaction filtering. |
+| end_date          | Optional[str]      | An optional ending date to your transaction filtering. |
+| category_ids      | List[str]          | An optional list of category IDs of transactions to include. |
+| tag_ids           | List[str]          | An optional list of tag IDs of transactions to include. |
+| descriptions      | List[str]          | An optional list of descriptions of transactions to include. |
+| account_ids       | List[str]          | An optional list of account IDs of transactions to include. |
+| match_all_filters | boolean            | Whether to match all supplied filters (True) or at least one (False) |
+| include_investment | boolean           | Whether to include those transactions that are associated with an Investment Account. |
+| remove_pending    | boolean            | Whether to remove those transactions that are still Pending. | 
+| limit             | int                | The page size of results. |
+| offset            | int                | The starting record of your results. |
+
+### Date Filters
+
+As mentioned above, the Date Filter is the date window for which to generate your trend analysis or for which to search transactions.  The supplied value must be one of the following enum values:
 
 | Enum Value | Description |
 | ---------- | ----------- |
@@ -185,7 +204,7 @@ As mentioned above, the Date Filter is the date window for which to generate you
 | 10         | All Time      |
 | 11         | Custom        |
 
-If you select a Custom Date Filter, then `start_date` and `end_date` are required fields.
+If you select a Custom Date Filter, then `start_date` and `end_date` are required fields.  Similarly, if you wish to use `start_date` and `end_date`, Custom Date Filter must be used.
 
 ### From Python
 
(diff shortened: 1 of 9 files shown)


eyadgaran (commented):
Nice! 

General thought about the abstraction: starting to see it break down on extension to previously unsupported routes. my intuition on that is because it conflates business/application logic with the underlying api structure. there is potential for shared code but because of the level in the inheritance structure it is becoming more like case statements than a clean interface.

unfortunately mint doesnt publish a client so that falls on you as well (and we all thank you for stepping up to give us one!), but separating those responsibilities into independent modules - imagine that one module was published by mint to manage internal api access and the outer layer was completely agnostic to how that was facilitated - would greatly simplify the code, making it more intuitive as well as more extensible. changes to mirror new updates to mint happen in one place with a clear contract, whereas user facing code focuses entirely on input structure (e.g. the cli, selenium browser) and data formatting (e.g. reformatting outputs, extending with analytical insights, etc) to make accessing and using mint data more useful

burkematthew:
@eyadgaran fair points!  The abstractions I made in the last few PRs were really meant to help us get this one implemented quicker without requiring too much of a refactor.  I was observing earlier while working on this that the `api.py` is starting to feel a bit large and so a refactor feels imminent, especially given the work that you started in your open PR.  Completely open to restructuring into independent modules.  That being said, the purpose of this PR in particular was simply to allow our users to start filtering Transactions.  I like to keep refactors as completely separate changes\PRs to help create a cleaner \clearer history of changes throughout the source.

burkematthew:
@mrooney just touching base on the review for this.

README.md
```diff
@@ -167,9 +167,28 @@ As mentioned above, the Report Type is the type of report for which to generate
...
+### Financial Data Transactions
 
-As mentioned above, the Date Filter is the date window for which to generate your trend analysis.  The supplied value must be one of the following enum values:
+If you want to provide a more granular filtering of your financial data transactions, you can select from a variety of search filters that are sent to Mint. 
+
+| Parameter         | Data Type          | Description  |
+| ----------------  | ------------------ | -----------  |
+| date_filter       | DateFilter.Options | The date window for which to filter your transactions. |
+| start_date        | Optional[str]      | An optional beginning date to your transaction filtering. |
+| end_date          | Optional[str]      | An optional ending date to your transaction filtering. |
```
mrooney, line 178: Could we include the date format in the descriptions for start/end so it's easy to understand how to use?

mintapi/api.py
```diff
@@ -520,9 +494,9 @@ def get_trend_data(
         date_filter : DateFilter.Options
             date window. must use predefined enum windows or pass a start and end date with CUSTOM enum
         start_date : Optional[str], optional
-            optional start date (YYYY-mm-dd) if using enum CUSTOM, by default None
+            optional start date (mm-dd-yy) if using enum CUSTOM, by default None
```
mrooney, line 497: Is this a breaking change?

mrooney (changes requested):
Thanks! Overall looks good just 2 minor questions.

mintapi/api.py
burkematthew: @mrooney it does, but I made a judgment call to ensure that we have consistent date formatting across the board with date inputs. I should've done this when we merged trends in the first place, but didn't. This way, we don't break transaction date input formats here.

README.md
burkematthew: @mrooney you got it!

mintapi/api.py
mrooney: @burkematthew got it. Is this a format that Mint uses in its output somewhere? That could make sense if so, otherwise I'd say mm-dd-yy is a bit Western-centric / ambiguous and would prefer YYYY-mm-dd.

burkematthew: @mrooney from what I have seen, I don't think it is. It's simply what we've used as accepted input historically in the took. That being said, I agree with your point (and typically prefer that date format anyways). However, it would be a breaking change that seems to me would be better suited for a major release. Where things stand now, at least we'd keep consistent date formatting for input.

mrooney: @burkematthew thanks for the explanation! Please excuse me if I'm being dense, but if it is a breaking change no matter what, and we both prefer the YYYY-mm-dd format, wouldn't it make more sense to have a breaking change to a better format instead of a breaking change to a format neither of us prefer? 🤔 If I misunderstood and the change as you have here is not a breaking change, then it makes sense to align now and improve later :)

burkematthew: @mrooney no worries! 

The difference is that we'd be breaking a format that has been around for quite some time (at least since I got involved) instead of breaking a format that has been around for a month and probably minimally adopted at this point.

mrooney: Got it, thanks for the context, I think your approach makes sense then. Since the two formats have different numbers of characters, we can later support both by detecting which one it is and document the preferred YYYY-mm-dd one.

mrooney (approved):
Thanks for your patience and discussion :D

PR #536, opened 2022-08-31
Fix SMS MFA by adding additional CSS selector

zsolteszku:
Fixing SMS MFA selection by adding `[data-testid="challengePickerOption_SMS_OTP"]` CSS selector.


mrooney:
Thanks! Seems safe since it is simply adding another available option. Is there a visually different button / UI that changed? Or just something under the hood?

mrooney:
Thanks again, got a confirmation of a fix so will merge!

mrooney (approved):

PR #538, opened 2022-09-23
New MFA email format

tsdg112:
Mint changed the email sent for MFA.  The MFA code is now in the subject line.


mrooney:
Thanks, I was going to ask if the old email could still be going out to anyone, so I like that it still supports that. I'll merge and release this :)

mrooney (approved):

Issue #540, opened 2022-09-24
Is initialize_serverless_chrome_driver missing?


CalvinChanCan:
Hi, I was going through the readme to setup mintapi on AWS Lambda. The readme mentioned there was a constructor for a pre-configured headless serverless chrome `initialize_serverless_chrome_driver` but I don't see this method available. Is this missing from the api?

mrooney:
Thanks for giving it a shot! I believe this is intended simply as a sort of stub method, ie `your_custom_chrome_driver_instantiation_here()`. From the linked page in that section, you can see they add a bunch of options before doing `driver = webdriver.Chrome(chrome_options=chrome_options)`. It's just showing how to pass a custom driver. If you have any suggestions to improve the README in this section, please do send them along via PR or suggestion!

PR #542, opened 2022-09-28
[FIX] Clarify that initialize_serverless_chrome_driver is not provided with mintapi

CalvinChanCan:
I was confused by the wording on the README that there was a preconfigured method that `initialize_serverless_chrome_driver` was provided but is actually a stub method as mentioned in #540 . So this PR is to clarify the language a bit to make it a bit more clear that this is not actually provided.


burkematthew (approved):
LGTM thanks!

mrooney:
Thanks, great clarification :)

Issue #547, opened 2022-10-08
Cannot get past Sign In screen with 2.8


jcsacto:
**Describe the bug**
I updated to 2.8 in my Dev environment and it can no longer get past the Sign in screen. 

What version of MintAPI are you using? 2.8
What command[s] did you run / steps to reproduce? 

import mintapi
import pandas as pd
import gspread
import time 
from df2gspread import df2gspread as d2g 
from oauth2client.service_account import ServiceAccountCredentials
from config import username, password

#WHEN CHROME DRIVER GETS OUT OF DATE, SIMPLY UPDATE CHROME AND UPDATE CHROME DRIVER IN THE MINTPIPELINE DIRECTOR

mint = mintapi.Mint(
    username,  # Email used to log in to Mint
    password,  # Your password used to log in to mint
    # Optional parameters
    mfa_method='soft-token',  # See MFA Methods section
                       # Can be 'sms' (default), 'email', or 'soft-token'.
                       # if mintapi detects an MFA request, it will trigger the requested method
                       # and prompt on the command line.
    mfa_input_callback=None,  # see MFA Methods section
                              # can be used with any mfa_method
                              # A callback accepting a single argument (the prompt)
                              # which returns the user-inputted 2FA code. By default
                              # the default Python `input` function is used.
    mfa_token='LSOFXXXXXXXXXXXXXXXXXXXXXXXXXXXX',   # see MFA Methods section
                      # used with mfa_method='soft-token'
                      # the token that is used to generate the totp
    intuit_account=None, # account name when multiple accounts are registered with this email.
    headless=True,  # Whether the chromedriver should work without opening a
                     # visible window (useful for server-side deployments)
                         # None will use the default account.
    session_path=r"C:\Users\Jon Crosby\Dropbox\Repos\python\mintPipeline\SessionPath", # Directory that the Chrome persistent session will be written/read from.
                       # To avoid the 2FA code being asked for multiple times, you can either set
                       # this parameter or log in by hand in Chrome under the same user this runs
                       # as.
    imap_account=None, # account name used to log in to your IMAP server
    imap_password=None, # account password used to log in to your IMAP server
    imap_server=None,  # IMAP server host name
    imap_folder='INBOX',  # IMAP folder that receives MFA email
    wait_for_sync=False,  # do not wait for accounts to sync
    wait_for_sync_timeout=300,  # number of seconds to wait for sync
	use_chromedriver_on_path=False,  # True will use a system provided chromedriver binary that
[...]

PR #549, opened 2022-10-14
[FIX] - Fix issue with sign in by updating CSS Selector.

CalvinChanCan:
There seems to be an issue #547 where users are having issues with signing into mint. It appears that mint has updated the class names in their CSS.

### DIFF
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -352,10 +352,11 @@ def sign_in(
         expected_conditions.presence_of_element_located(
             (
                 By.CSS_SELECTOR,
-                "ius-hosted-ui-main-container",
+                ".ius-hosted-ui-main-container, #ius-link-use-a-different-id-known-device, #ius-userid, #ius-identifier, #ius-option-username",
             )
         )
     )
+
     driver.implicitly_wait(0)  # seconds
 
     user_selection_page(driver)
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -352,7 +352,8 @@ def sign_in(
         expected_conditions.presence_of_element_located(
             (
                 By.CSS_SELECTOR,
-                ".ius-hosted-ui-main-container, #ius-link-use-a-different-id-known-device, #ius-userid, #ius-identifier, #ius-option-username",
+                ".ius-hosted-ui-main-container, #ius-link-use-a-different-id-known-device, #ius-userid, "
+                '#ius-identifier, #ius-option-username, [data-testid="IdentifierFirstSubmitButton"]',
             )
         )
     )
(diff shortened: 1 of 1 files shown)


mintapi/signIn.py
```diff
@@ -352,7 +352,7 @@ def sign_in(
         expected_conditions.presence_of_element_located(
             (
                 By.CSS_SELECTOR,
-                "#ius-link-use-a-different-id-known-device, #ius-userid, #ius-identifier, #ius-option-username",
+                "ius-hosted-ui-main-container",
```
burkematthew, line 355: Given that there's a good chance not all users have this new selector, I think we should just append to the existing list instead of starting a new one with just this one value.

burkematthew (changes requested):

mintapi/signIn.py
CalvinChanCan: Thanks. I updated the CSS Selector to support the old CSS Class Names. However, for the new one to work for me, I had to use ".ius-hosted-ui-main-container" rather than "#ius-hosted-ui-main-container". I'm unsure if that's how the class names selectors worked previously.

snowskeleton:
None of these potential changes seem to get me past the login screen. I've tried replacing with/appending the new selector, with/without the hash/period, and I'm still left staring at my username until it crashes. I'm on chrome v 106.

tsdg112:
They also changed the CSS selector for the bypass "Let's add your current mobile number" interstitial page (at line 534).  The following changes worked for me:

```
355c355
<                 "#ius-link-use-a-different-id-known-device, #ius-userid, #ius-
identifier, #ius-option-username",
---
>                 '#ius-link-use-a-different-id-known-device, #ius-userid, #ius-
identifier, #ius-option-username, [data-testid="IdentifierFirstSubmitButton"]',
442c442
<                 (By.CSS_SELECTOR, "#ius-userid, #ius-identifier, #ius-option-u
sername")
---
>                 (By.CSS_SELECTOR, '#ius-userid, #ius-identifier, #ius-option-u
sername, [data-testid="IdentifierFirstSubmitButton"]')
534c534,535
<             By.ID, "ius-verified-user-update-btn-skip"
---
>             By.CSS_SELECTOR,
>             '#ius-verified-user-update-btn-skip, [data-testid="VUUSkipButton"]
',

```

snowskeleton:
I figured out why I was still having issues. These changes fix the problem if I don't include the session file from previous login attempts, presumably because it takes me through a different login flow. I'm pretty sure it's the `handle_different_page_username_password` function that's breaking for me, but idk exactly where.

CalvinChanCan:
Thanks @tsdg112 . I've added your CSS selectors to my PR and gave you credit in the commit. 

@snowskeleton , do you have the python stack trace when you encounter the error?

tsdg112:
@CalvinChanCan: Thank you for including my changes.

snowskeleton:
> @snowskeleton , do you have the python stack trace when you encounter the error?

I just went back to reproduce it, but the issue appears to be gone with he latest commit. hurray!

burkematthew (approved):

Issue #545, opened 2022-10-08
Cannot log in using canonical use case example


felciano:
**Describe the bug**
I have just done a clean install of `mintapi` v2.8, and am unable to get the basic CLI example to work. Here's what I'm calling:
```
poetry run mintapi --keyring <myusername>
```
It opens the Selenium browser and correctly manages the username and password pages. However when the MFA page opens, the value is not filled in, nor am I prompted for it on the command line. After a little bit, it fails out with a `Message: invalid selector: No selector specified` error.

Using Chrome 106.0.5249.103 and python 3.10.7

**What did you expect to happen?**
Expected `mintapi` to prompt me for my MFA code.

**What actually happened?**
Nothing initially, then exceptioned out.

**Additional context**
Full stack trace below:
```
Message: invalid selector: No selector specified
  (Session info: chrome=106.0.5249.103)
Stacktrace:
0   chromedriver                        0x0000000102aa8598 chromedriver + 4404632
1   chromedriver                        0x0000000102a2ffa3 chromedriver + 3911587
2   chromedriver                        0x00000001026dce6a chromedriver + 425578
3   chromedriver                        0x00000001026e0037 chromedriver + 438327
4   chromedriver                        0x00000001026dfed7 chromedriver + 437975
5   chromedriver                        0x00000001026e019c chromedriver + 438684
6   chromedriver                        0x00000001027149a0 chromedriver + 653728
7   chromedriver                        0x0000000102714e91 chromedriver + 654993
8   chromedriver                        0x000000010274c274 chromedriver + 881268
9   chromedriver                        0x0000000102732dcd chromedriver + 777677
10  chromedriver                        0x0000000102749f79 chromedriver + 872313
11  chromedriver                        0x0000000102732bd3 chromedriver + 777171
12  chromedriver                        0x0000000102707ced chromedriver + 601325
13  chromedriver                        0x0000000102708e51 chromedriver + 605777
14  chromedriver                        0x0000000102a780ce chromedriver + 4206798
15  chromedriver                        0x0000000102a7c797 chromedriver + 4224919
16  chromedriver                        0x0000000102a83a4f chromedriver + 4254287
17  chromedriver                        0x0000000102a7d4da chromedriver + 4228314
18  chromedriver                        0x0000000102a5623f chromedriver + 4067903
19  chromedriver                        0x0000000102a9a6a8 chromedriver + 4347560
20  chromedriver                        0x0000000102a9a816 chromedriver + 4347926
[...]

Issue #546, opened 2022-10-08
WebDriverException crash after successful log in with v2.8


felciano:
**Describe the bug**
After successfully logging in, `mintapi` v2.8 crashes out with a Selenium error.

```
poetry run mintapi --keyring --mfa-method=soft-token --mfa-token <my-soft-token> <my-username>
```

Using Chrome 106.0.5249.103 and python 3.10.7

**What did you expect to happen?**
Default behavior with no parameters (account data returned).

**What actually happened?**
The chromedriver instance logs in correctly (username, password and MFA) and the main UI comes up. It sites for a few seconds, then `mintapi` exceptions out.

**Additional context**
Trace is below:
```
Traceback (most recent call last):
  File "/Users/felciano/.cache/pypoetry/virtualenvs/mintapitest-b8xIeIvi-py3.10/bin/mintapi", line 8, in <module>
    sys.exit(main())
  File "/Users/felciano/.cache/pypoetry/virtualenvs/mintapitest-b8xIeIvi-py3.10/lib/python3.10/site-packages/mintapi/cli.py", line 518, in main
    data = mint.get_account_data(limit=options.limit)
  File "/Users/felciano/.cache/pypoetry/virtualenvs/mintapitest-b8xIeIvi-py3.10/lib/python3.10/site-packages/mintapi/api.py", line 262, in get_account_data
    return self.get_data(
  File "/Users/felciano/.cache/pypoetry/virtualenvs/mintapitest-b8xIeIvi-py3.10/lib/python3.10/site-packages/mintapi/api.py", line 242, in get_data
    data = self.__get_mint_endpoint(endpoint, **kwargs)
  File "/Users/felciano/.cache/pypoetry/virtualenvs/mintapitest-b8xIeIvi-py3.10/lib/python3.10/site-packages/mintapi/api.py", line 449, in __get_mint_endpoint
    response = self.get(
  File "/Users/felciano/.cache/pypoetry/virtualenvs/mintapitest-b8xIeIvi-py3.10/lib/python3.10/site-packages/mintapi/api.py", line 162, in get
    return self.driver.request("GET", url, **kwargs)
  File "/Users/felciano/.cache/pypoetry/virtualenvs/mintapitest-b8xIeIvi-py3.10/lib/python3.10/site-packages/seleniumrequests/request.py", line 159, in request
    self.requests_session.headers = get_webdriver_request_headers(self, proxy_host=self.__proxy_host)
[...]

PR #551, opened 2022-10-14
[FIX] Fix issue with user selection page by updating CSS Selector

CalvinChanCan:
As @snowskeleton mentioned in #549 , there is also a sign in issue with the updated CSS in the user selection page. The issue can be recreated by passing your own webdriver but using your own Chrome User Profile that has been previously logged in. Since mint already has a session saved, it should go through the user selection page flow and then raise an error for not finding "Use a different user ID" button.

This PR should address this issue for anyone else going through the user selection page flow.

A side note, but the general new CSS selector trend seems to be `data-testid=`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ __pycache__
 /.idea
 .vscode/
 chromedriver*
+.venv
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
     New feature work occurs in the 2.x release series and only supports the new Mint UI / endpoints.
     Please see the README for more information.
     
+2.9
+---
+- various fixes to CSS selectors for login (#549, #552)
+
 2.8
 ---
 - handle new MFA email format w/ code in subject (#538)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 --> NOTE! The 1.x release series only supports and contain fixes for the previous Mint UI.
     New feature work occurs in the 2.x release series and only supports the new Mint UI / endpoints.
     Please see the README for more information.
+    
+2.8
+---
+- handle new MFA email format w/ code in subject (#538)
+- update for new SMS MFA UI (#536)
+- improve README for creating and passing in your own driver (#542)
 
 2.7
 ---
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -72,9 +72,9 @@ ENDPOINTS = {
 }
 
 
-def convert_mmddyy_to_datetime(date):
+def convert_mmddyy_to_yyyymmdd(date):
     try:
-        newdate = datetime.strptime(date, "%m/%d/%y")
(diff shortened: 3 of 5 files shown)


CalvinChanCan:
@felciano thanks for testing this! When I was using my Chrome User Profile I had thought Chrome had been just autofilling my username. However, I believe that mint now populates the email input field already with the email from the session. But for some reason `email_input.clear()` is not actually clearing the value in the email input and so we get the email inputted twice. So my workaround now uses a loop to send backspace keys the length of the email to remove the email autofilled by mint.

burkematthew:
@CalvinChanCan have you tried waiting for the element to be clickable to make sure that we can actually click to clear?

https://stackoverflow.com/a/49151715

CalvinChanCan:
@burkematthew Thanks! The waiting for the `email_input` to be clickable worked! I added in the WebDriverWait and changed it back to clear the input rather than the workaround with the backspaces.

CalvinChanCan:
@burkematthew So I took a closer look and had thought waiting for the email input and then attempting to cleared it work. But I accidentally removed `email_input.send_keys(email)` so it had been using the email address from the session already.

I tried waiting for the email input, then attempted to add the email to the email but its showing up as "USERNAME USERNAME" again. From the stackoverflow link, I tried adding a click before inputting the email but still getting not able to successfully clear the email.

```
        email_input = WebDriverWait(driver, 20).until(
            expected_conditions.element_to_be_clickable(
                (
                    By.CSS_SELECTOR,
                    '#ius-identifier, [data-testid="IdentifierFirstIdentifierInput"]',
                )
            )
        )
        if not email_input.is_displayed():
            raise ElementNotVisibleException()

        email_input.click()
        email_input.clear()
        email_input.send_keys(email)
```

felciano:
I tried `1d8ce4e` and it is now completing the login sequence as expected (username, password and MFA), but then fails out on a
```
selenium.common.exceptions.WebDriverException: Message: disconnected: received Inspector.detached event
```
which was not happening before. Are you seeing this as well? This is with a default `mintapi` call using soft-tokens:

```
poetry run mintapi --keyring --mfa-method=soft-token --mfa-token "SOFTTOKEN" "USERNAME"
```

felciano:
@CalvinChanCan actually I'm fairly certain that the above `received Inspector.detached event` is #546 and independent of the login issue this patch is trying to address. From my perspective, this patch successfully re-enables automated login and is good to go

CalvinChanCan:
Hmm I'm not quite sure this patch is working as intended. It certainly is getting through the user selection page but its odd that it does when I accidentally left out `email_input.send_keys(email)`. So what I think is happening is that mint is still using the email from the session rather than the email we're passing into the mintapi constructor. 

I have tried a few ways of trying to get the email to successfully clear but haven't found anything that works besides the backspace method. @burkematthew , do you have any thoughts if we should revert back to the backspace method or what we can do to resolve this issue?

felciano:
I believe this patch fixes #545 (referenced here to track it)

CalvinChanCan:
I'll close this PR since this does not seem to be an issue.

PR #552, opened 2022-10-15
Add additional CSS selectors for signin process

apiology:
The current set does not match what I now receive on https://accounts.intuit.com/index.html:

image

### DIFF
--- a/mintapi/signIn.py
+++ b/mintapi/signIn.py
@@ -575,6 +575,18 @@ def bypass_passwordless_login_page(driver):
         ElementNotInteractableException,
     ):
         pass
+    # bypass "Let's make sure you're you" if password login is allowed
+    try:
+        skip_for_now = driver.find_element(
+            By.CSS_SELECTOR, '[data-testid="challengePickerOption_PASSWORD"]',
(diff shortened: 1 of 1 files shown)


apiology:
image

Another CSS fix in https://github.com/mintapi/mintapi/pull/552/commits/9af80cc15895b50b12b8c6b5527bcd32cf3bf74b for this page, which reproduced for me in headless mode.

apiology:
https://github.com/mintapi/mintapi/pull/552/commits/d7d57d7a579c8dde7993077ca6af7027aab6e67f overcomes this page, which seems to offer the user the choice of password or a different factor: image

apiology:
Verified that the patched mintapi works all the way through again in the custom Docker image I typically run mintapi in, though I'm experiencing #546 and #553 when running locally on macOS.

apiology:
If folks want to try out my branch, this will do it: `pip uninstall mintapi && pip install git+https://github.com/apiology/mintapi.git@account_selection_css_alternative`

mrooney (approved):
Thanks for your help, looks good to me!

Issue #556, opened 2022-10-19
get_transaction_data's date filters being ignored and return all data.


CalvinChanCan:
**Describe the bug**
When using a date filter to get transactions between a certain date, `get_transaction_date` is returning all transactions instead of only the transactions between the specified dates.

What version of MintAPI are you using? Main `17ae070``

What command[s] did you run / steps to reproduce?

```
    client = mintapi.Mint(username, password)
    data = client.get_transaction_data(start_date='01/01/18', end_date='12/31/18')
```

Even trying the DateFilter Option Enum generates an error.

Stacktrace/error received:
When running with the DateFilter Option Enum as Custom
```
    client = mintapi.Mint(username, password)
    data = client.get_transaction_data(date_filter=DateFilter.Options.CUSTOM, start_date='01/01/18', end_date='12/31/18')

```
```
Traceback (most recent call last):
  File "\mintapi\main.py", line 14, in <module>
    data = client.get_transaction_data(date_filter=DateFilter.Options.CUSTOM, start_date='01/01/18', end_date='12/31/18')
  File "\mintapi\mintapi\api.py", line 603, in get_transaction_data
    data = self.get_data(constants.POST_METHOD, constants.TRANSACTION_KEY, payload)
  File "\mintapi\mintapi\api.py", line 240, in get_data
    data = self.__post_mint_endpoint(endpoint, payload)
  File "\mintapi\mintapi\api.py", line 459, in __post_mint_endpoint
    response = self.post(
  File "\mintapi\mintapi\api.py", line 165, in post
    return self.driver.request("POST", url, **kwargs)
  File "\mintapi\venv\lib\site-packages\seleniumrequests\request.py", line 216, in request
    response = self.requests_session.request(method, url, **kwargs)
  File "\mintapi\venv\lib\site-packages\requests\sessions.py", line 573, in request
    prep = self.prepare_request(req)
  File "\mintapi\venv\lib\site-packages\requests\sessions.py", line 484, in prepare_request
    p.prepare(
  File "\mintapi\venv\lib\site-packages\requests\models.py", line 371, in prepare
    self.prepare_body(data, files, json)
  File "\mintapi\venv\lib\site-packages\requests\models.py", line 511, in prepare_body
    body = complexjson.dumps(json, allow_nan=False)
  File "C:\Program Files\Python39\lib\json\__init__.py", line 234, in dumps
    return cls(
  File "C:\Program Files\Python39\lib\json\encoder.py", line 199, in encode
[...]

PR #557, opened 2022-10-19
[FIX] - Issue with date filters being ignored

CalvinChanCan:
This PR should address issue #556 where the date filters are being ignored. The format on mint has the start and end dates a string in the format of "YYYY-MM-DD" but mintapi had been passing them as datetime objects which does not serialize when converting to a dictionary.

This can be test manually providing the DateFilter.Options.CUSTOM as the date_filter and start and end date. There should not be any transactions outside of the provided date range.

For example:
```
    client = mintapi.Mint(username, password)
    data = client.get_transaction_data(date_filter=DateFilter.Options.CUSTOM, start_date='01/01/18',
                                       end_date='12/31/18')
```

### DIFF
--- a/mintapi/api.py
+++ b/mintapi/api.py
@@ -74,7 +74,7 @@ ENDPOINTS = {
 
 def convert_mmddyy_to_datetime(date):
     try:
-        newdate = datetime.strptime(date, "%m/%d/%y")
+        newdate = datetime.strptime(date, "%m/%d/%y").strftime("%Y-%m-%d")
     except (TypeError, ValueError):
         newdate = None
     return newdate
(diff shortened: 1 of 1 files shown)


mintapi/api.py
```diff
@@ -649,8 +649,8 @@ def __build_payload(
         return request(
             date_filter=DateFilter(
                 date_filter=date_filter,
-                start_date=convert_mmddyy_to_datetime(start_date),
-                end_date=convert_mmddyy_to_datetime(end_date),
+                start_date=convert_mmddyy_to_datetime(start_date).strftime("%Y-%m-%d"),
```
burkematthew, line 652: I'd prefer if we just update the method so we're ways returning the date in its required format.

burkematthew (changes requested):

burkematthew (approved):

PR #561, opened 2022-10-24
Add quit_driver_on_fail flag

zsolteszku:
Add the option to not quit the `webdriver` when an exception happens. Usecase: when there is an exception through the login process I would like to take a screenshot and save the HTML file for debugging purposes but I cannot as `Mint` always closes the driver.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,9 +245,11 @@ make calls to retrieve account/budget information.  We recommend using the
     wait_for_sync=False,  # do not wait for accounts to sync
     wait_for_sync_timeout=300,  # number of seconds to wait for sync
     fail_if_stale=True, # True will raise an exception if Mint is unable to refresh your data.
-	use_chromedriver_on_path=False,  # True will use a system provided chromedriver binary that
+	  use_chromedriver_on_path=False,  # True will use a system provided chromedriver binary that
 	                                 # is on the PATH (instead of downloading the latest version)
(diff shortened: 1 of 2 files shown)


mrooney:
Given the use-case, would it be more useful to provide a flag to grab a screenshot and HTML file before quitting? Is that something that can be automated?

zsolteszku:
> Given the use-case, would it be more useful to provide a flag to grab a screenshot and HTML file before quitting? Is that something that can be automated?

IMHO if the driver is passed to the Mint constructor, Mint object should not take ownership of it (like closing the driver) as the ownership probably belongs to the caller. For my usecase yes I wanted to take a screenshot but with this flag the caller can do whatever when login fails. 

I am not sure that the right approach is to support this by default, but if there is a strong preference to support it I can take a look. In terms of how easy is to support this to take a screenshot by default when there is a login problem: not that hard, need to change some chrome options about the download folder but not sure how to access the screenshot if it runs through docker.

mrooney:
Thanks, sounds good! Would you mind just adding it to the README so folks know this is an option when creating a driver?

burkematthew (changes requested):
I second @mrooney 's request for an addition to the README. I'd also like to see this added as an option to the CLI so users that leverage that method can also use this new feature. Thanks for your contribution!

burkematthew (approved):

PR #568, opened 2022-11-15
prevent email MFA check from marking messages as read

tsdg112:
The fixes a problem that happens when checking email for the MFA code.  When it reads non-MFA messages, they would be marked as read.  This fixes that problem.

It also adds a check for an additional cause of sign-in failure: when you are temporarily blocked by Mint.


mrooney:
Thanks! LGTM to me on both accounts but would also love a review from @burkematthew as I'm not nearly as familiar with the email flow.

burkematthew:
@mrooney I'll try and take a look at this this weekend. If it looks good, I'll approve as well and then merge.

tsdg112:
Mint made a minor change to the email subject.  The latest commit allows both the old and new subject.

mrooney:
Thanks for the update with both versions :) @burkematthew if you happen to get a minute, I would love to know if all looks well to you.

burkematthew (approved):

PR #578, opened 2022-12-04
Updated duplicate of #482 [REST API]

snowskeleton:
@eyadgaran  did amazing work on PR #482 . This PR brings catches up with upstream changes.


burkematthew:
@snowskeleton any chance this can be implemented in a backwards compatible way? Meaning we have a way to support both the new RESTful approach and the original approach. I'd be more inclined to get this in sooner if we could have that. Then, we give people plenty of time to acclimate to the new approach and in the next major release, we officially deprecate the original approach.

snowskeleton:
Theoretically yes. It's setup in such a way that any endpoints without a REST implementation fallback to Selenium. It should be easy enough to add a flag that gives the user control over which style to use. Maybe for this version REST is opt-in, and next major release REST is default with the ability to fallback.

At the moment I'm taking a look at the failing tests. Gotta make those angry X's go away

eyadgaran:
thanks for picking this up @snowskeleton! i have a couple hours this weekend that i can use to help out. i'll start with the suggestions in the thread - let me know what else would be helpful

mrooney:
Awesome! That PR looks very promising. I'd certainly be eager to integrate it in a way that doesn't break existing usage, CLI and via Python. I like the idea of a flag that can be used to opt-in, and we can document that this would become the default method in 3.0, if it makes sense.

mrooney:
Closing in favor of #482

Issue #567, opened 2022-11-14
Enhancement: Several Docker builds pinned to specific recent chrome/chromedriver versions


felciano:
This issue is to track a new feature (or capability) request for `mintapi` that might provide some buffer against future chrome-based issues. It looks like there is a Docker configuration for `mintapi` now that allows you to run it in an isolated environment. However I assume this includes the most recent versions of chrome/chromedriver, which sometimes results in breakage.

So the idea here is to start creating multiple Docker builds, each with the latest mintapi code but with different recent versions of chrome/chromedriver. So at any point we might pick from e.g. the last 3 versions:

- docker_mintapi_2.9_chrome_105
- docker_mintapi_2.9_chrome_106
- docker_mintapi_2.9_chrome_107

I haven't explored Docker myself so no idea if this is straightforward or not, but I wanted to get the discussion going to see if this seems like a good idea.

snowskeleton:
I like the idea behind this.

Currently with Docker, the chrome and chromedriver versions are up to Selenium. I imagine we could specify a Selenium version and that should prevent unintended upgrades.

mrooney:
This seems like a good idea!

burkematthew:
Instead of multiple Dockerfile, I think we could potentially have an argument based base image. That way we don't have to have ongoing upkeep of the different ones and it provides the flexibility for users to control which base to use. That being said, I haven't done enough research to know how viable it is.

snowskeleton:
If we want some level of stability increase with [selenium](https://github.com/SeleniumHQ/docker-selenium), an easy solution would be to find a version that works and just stick to that version. For example with the current release, the line in `Dockerfile` would be changed from
```
FROM selenium/standalone-chrome
```
to
```
FROM selenium/standalone-chrome:4.7.0
```
This would get us halfway to the utopia of not having to worry about chrome versions, with the caveat being that the version chosen has to be solid. Then every few months we can bump the version, after we verify it works.

burkematthew:
@snowskeleton yeah I like that approach too.

PR #580, opened 2022-12-08
Update Dockerfile

snowskeleton:
per issue #567 . this locks Selenium to use chrome and chromebrowser 108.


mrooney:
Seems reasonable to me given all the back and forth issues. Any concerns @burkematthew

mrooney:
Just reviewing PRs and wanted to check in on this one. Do you think the more future proofed selenium-requests resolves this, or would you still be a fan of this approach? I could see arguments either way. If we like it, would it make sense to bump it to the latest working versions?

snowskeleton:
I'm not familiar with what measures have been taken in `selenium-requests` to resolve this issue.
In general, though, it's usually best not to rely on another source to prevent your stuff from breaking, so I think it makes sense to pin the version on our end.

As for bumping it to the latest working version, as long as it works I can't think of a reason not to.

PR #583, opened 2023-01-11
Rework MFA section

snowskeleton:
lead the user down the happy path.


mrooney:
Very awesome, thank you! I'm curious if this should even go above Usage, if everyone is going to need to see this section. What do you think? Otherwise looks good and this is an awesome change :)

snowskeleton:
The worst case scenario for first-time users who don't scroll down enough to read the reworked section is they might have to manually copy an MFA code, which isn't all that onerous.
Also, my intuition says the only users likely to care are those who will go through the whole `README`, so placement is irrelevant.
For that reason, I don't believe the increased visibility is worth the cost of increased screen clutter and extra scrolling for new users.

I like the idea of reworking the Automation section, perhaphs with the MFA bit near the top, which would put MFA right below Usage. If you'd like, I'll work on the layout a bit and submit that as a separate PR.

mrooney:
Thanks, seems fair. I just wasn't sure if everyone was going to need this section. I'll merge this and a follow up PR sounds great :)

mrooney (approved):

README.md
```diff
@@ -101,13 +101,48 @@ mint = mintapi.Mint(..., driver=driver)
...
+but Mint will sometimes **still email you a second factor code**.
+So, for the least fragility, enable MFA.
+
+The `mfa_method` parameter is only required if your login flow presents you with the option
+to select which Multifactor Authentication Method you wish to use,
+typically as a result of your account configured to accept different methods.
+Prior to v2.0, `mfa_method` is always required.
+
+- #### TOTP
+Set `mfa_method` to `soft-token`.
```
hudcap, line 124: This is in apparent contradiction to the previous paragraph, however, I was only able to get TOTP to work by setting the `mfa_method` to `soft-token`, so I think the previous paragraph should be modified. (Longer discussion in #484)

mrooney: @hudcap maybe it could be modified to state that it is best practice to try without mfa_method first, then specify one if it doesn't work, and open a bug report? Let me know if you have thoughts on this specific case @burkematthew

snowskeleton: I think there's a difference in application. By my understanding, setting `mfa_method` is only required past v2.0 for automated deployment, but not where `headless=False` during setup and/or troubleshooting. It's only required for automation because you don't want to sit there and manually enter the TOTP. (If this is wrong, please correct me.)

I'm not sure the best way to go about wording that in the README. If we expect most users to want automated deployment, we should default to simply saying that `mfa_method` is required. If we want to make the first-time startup instructions quicker/easier, I vote for leaving it marked as optional with a note along the lines of "required for automated use".

thoughts?

burkematthew: We enhanced `mintapi` so that it detects which mfa method is requested by Mint.  It's been awhile since I've looked at it and have frankly never used a soft token.  So I can't remember _why_ it's necessary to specify.  The wording should basically indicate that unless you're using a `soft_token`, don't specify the `mfa_method`.  That being said, it's possible there's some edge casing and cleanup that could be done in the `soft_token` scenario so that it's a more uniform experience around MFA.

FWIW, using the email mfa_method allows for automation since you can use IMAP to connect with your inbox, which is how I've automated my implementation.

PR #584, opened 2023-01-12
Ignore `LICENSE.chromedriver` file

snowskeleton:
I just had these start showing up in my project folder.


mrooney:
Thanks! Since this is one specific file, could we just add a line for it explicitly?

snowskeleton:
how's that look? I also moved it up with the other `chromedrive` ignore

mrooney:
Nice, thanks :D

mrooney (approved):

mrooney:
I would think `chromedriver*` makes `chromedriver` redundant but I'm not positive. Of course that wasn't introduced in this PR so it's strictly better now :)

PR #587, opened 2023-01-31
README update

snowskeleton:
Bit of a makeover. I tried to preserve as much information as possible while condensing and reorganizing, but lemme know if I missed anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
     New feature work occurs in the 2.x release series and only supports the new Mint UI / endpoints.
     Please see the README for more information.
     
+2.13
+---
+- use a RESTful API architecture instead of heavily relying on Selenium (#482)
+  - you can opt-in to full use of this via `--use-rest-client` (CLI) or `use_rest_client=True` (Python)
+- improved detection of refresh status (#583)
+
 2.12
 ---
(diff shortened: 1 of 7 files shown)


mrooney (changes requested):
Looks great, thanks for the re-work! A couple small requests:

* Let's keep the Discord at the top unless there's a convincing reason to move it to the bottom. I suspect a lot less people will find it there.
* For the MFA section, it says to only set mfa-method if you are presented with multiple options, but then says to use it for TOTP in the next section. Would this be clearer as "As of v2.0, the mfa_method parameter is only required when using soft-token, or if your login flow presents..."?
* "Best practice is to pass dedicated chrome and chromedriver binaries to `mintapi`." Does this actually mean specify a custom `PATH`, or do you just mean using the argument to use Chrome on the PATH?

snowskeleton:
last two requested changes were easy enough.
For the discord info, how do you like the placement now? I don't think it makes sense to have it above the basic setup info, but here it's still close to the top and thus more discoverable.

mrooney (changes requested):
Thanks, looks awesome! The only thing I notice is one black error (it seems odd that there would be black issues on unchanged files) and also one typo of "Selinium".

mrooney (approved):
Awesome, thank you!

Issue #588, opened 2023-02-18
Headless mode is changing in selenium


ddollar:
https://www.selenium.dev/blog/2023/headless-is-going-away/

Looks like `--headless` needs to be changed to `--headless=new`

I'm on chromium/chromedriver `110.0.5481.77` (from apt in Debian 11)

https://github.com/SeleniumHQ/selenium/issues/11634

gyllen:
I second that switching the parameter to "headless=new" works for the newer versions of chromium.

mrooney:
Thanks! If anyone is interested in making a PR with this change as well as the relevant dependency changes to use the new versions, that would be great!

gyllen:
Created a pull request but as mentioned in note, took the road of just adding the flag and not changing the old one, needs to be tested with older versions of chromium -> https://github.com/mintapi/mintapi/pull/589

apiology:
I gather this is the symptom for this problem:

```
  File "/usr/bin/mintapi", line 8, in <module>
    sys.exit(main())
  File "/usr/lib/python3.10/site-packages/mintapi/cli.py", line 465, in main
    mint = Mint(
  File "/usr/lib/python3.10/site-packages/mintapi/api.py", line 124, in __init__
    self.login_and_get_token(
  File "/usr/lib/python3.10/site-packages/mintapi/api.py", line 192, in login_and_get_token
    self.driver = driver or _create_web_driver_at_mint_com(
  File "/usr/lib/python3.10/site-packages/mintapi/signIn.py", line 311, in _create_web_driver_at_mint_com
    driver = Chrome(options=chrome_options)
  File "/usr/lib/python3.10/site-packages/seleniumrequests/request.py", line 148, in __init__
    super(RequestsSessionMixin, self).__init__(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/selenium/webdriver/chrome/webdriver.py", line 80, in __init__
    super().__init__(
  File "/usr/lib/python3.10/site-packages/selenium/webdriver/chromium/webdriver.py", line 104, in __init__
    super().__init__(
  File "/usr/lib/python3.10/site-packages/selenium/webdriver/remote/webdriver.py", line 286, in __init__
    self.start_session(capabilities, browser_profile)
  File "/usr/lib/python3.10/site-packages/selenium/webdriver/remote/webdriver.py", line 378, in start_session
    response = self.execute(Command.NEW_SESSION, parameters)
[...]

PR #590, opened 2023-02-23
Add new chromium flag for headless

gyllen:
Add the new headless=new flag for headless browsing in chromium.

NOTE: Took the approach of just adding the flag, not backtested with old chromium versions (hopefully old versions ignore bogus flags)


apiology:
This fix resolved the issue for me.

apiology:
Fixes #588

burkematthew:
I'd like to see us check the chromium version used and based on that, set the appropriate headless option.

Issue #592, opened 2023-03-01
AttributeError: 'NoneType' object has no attribute 'execute_script'


ktorq:
It seems something has broke between 2.12 and 2.13 - the below code worked fine on 2.12
```
client = mintapi.Mint(keys['mint_user'], keys['mint_pw'], headless=False, wait_for_sync=False, mfa_method='authenticator')
transactions = pd.DataFrame(client.get_transaction_data(remove_pending=False))
transactions.to_csv('mint_transactions.csv')
```

but if I upgrade to 2.13 I get this error:
```
Traceback (most recent call last):
  File "C:/main.py", line 109, in <module>
    main()
  File "C:/main.py", line 102, in main
    get_mint_transactions()
  File "C:/main.py", line 36, in get_mint_transactions
    transactions = pd.DataFrame(client.get_transaction_data(remove_pending=False))
  File "C:\venv\lib\site-packages\mintapi\endpoints.py", line 575, in get_transaction_data
    data = self._get_transaction_data(json=payload.to_dict(), **kwargs)
  File "C:\venv\lib\site-packages\mintapi\endpoints.py", line 283, in _get_transaction_data
    return self.post(
  File "C:\venv\lib\site-packages\mintapi\endpoints.py", line 37, in post
    return self.request(method="POST", **kwargs)
  File "C:\venv\lib\site-packages\mintapi\browser.py", line 184, in request
    auth_headers = self._get_api_key_header()
  File "C:\venv\lib\site-packages\mintapi\browser.py", line 212, in _get_api_key_header
    api_key = self.driver.execute_script("return " + key_var)
AttributeError: 'NoneType' object has no attribute 'execute_script'
```

I also noticed that for some reason 'authenticator' was removed as a possible mfa-method. Not sure why this was removed as it worked perfectly fine for me pre-2.13. I tried mfa_method='soft-token' and tried removing the mfa_method altogether but still get the above error so not related but just pointing it out.

ktorq:
Here is the new code for 2.13:

https://github.com/mintapi/mintapi/blob/c00def85b34e72ba099988e4ee35c02dbc024f71/mintapi/browser.py

```
 def _get_api_key_header(self):
        key_var = "window.__shellInternal.appExperience.appApiKey"
        api_key = self.driver.execute_script("return " + key_var)
        auth = "Intuit_APIKey intuit_apikey=" + api_key
        auth += ", intuit_apikey_version=1.0"
        header = {"authorization": auth}
[...]

PR #593, opened 2023-03-28
[CHORE] - Update README with 2.13 breaking changes and code examples

CalvinChanCan:
This PR address the issue mentioned in https://github.com/mintapi/mintapi/issues/592 where user's code is breaking due to the constructor in 2.13 now using keyword arguments instead of positional arguments for email and password. I've also update the code examples for 2.13 and examples to maintain backwards compatibility using the legacy browser based client. 

This PR is a draft and could use some feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,8 +260,8 @@ make calls to retrieve account/budget information.  We recommend using the
 ```python
   import mintapi
   mint = mintapi.Mint(
-    'your_email@web.com',  # Email used to log in to Mint
-    'password',  # Your password used to log in to mint
+    email='your_email@web.com',  # Email used to log in to Mint
+    password='password',  # Your password used to log in to mint
     # Optional parameters
     mfa_method='sms',  # See MFA Methods section
                        # Can be 'sms' (default), 'email', or 'soft-token'.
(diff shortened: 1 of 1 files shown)


mrooney:
Thanks this is awesome! I don't believe that was intended to be a breaking change, maybe this PR could address that my making it backwards compatible. @eyadgaran @snowskeleton do you happen to know if that was intended and if there is a clean way we could adjust the code to work for those scenarios?

CalvinChanCan:
@mrooney, after more testing, I realized that the only breaking change was actually the positional arguments vs keyword arguements. The other issue I mentioned is not actually an issue since `__getattr__` will just route between the browser client or rest_client. Though this does not give any method hints. https://github.com/mintapi/mintapi/blob/main/mintapi/api.py#L52

burkematthew (approved):

eyadgaran:
thanks - definitely no breaking changes intended! must've missed the usage. re the type hints the dynamic routing was an interim solution to support both clients (browser and rest) and can change to one in a future major version change with proper inheritance

mrooney:
That's awesome and thanks for the tests! I'll merge and release this as 2.14 :)

PR #595, opened 2023-04-07
[CHORE] - Update Mint constructor tests

CalvinChanCan:
This PR updates the Mint constructor tests for both the rest client and browser client when using positional arguments and keyword arguments.

It tests the 4 different possibilities for passing the username and email into the mint constructor are:

Keyword arguments using the rest client
Positional arguments using the rest client
Keyword arguments using the browser client
Positional arguments using the browser client

My previous PR which added these tests were not testing all four possibilities.


burkematthew (approved):

mrooney:
Awesome, thanks so much!

PR #596, opened 2023-04-11
[FEATURE] - Add endpoint to update transaction data

CalvinChanCan:
Not sure if this is allowed and feel free to close if not.

This PR adds an endpoint to update an existing transaction with an updated description, notes, category, and tags and to retrieve the updated transaction.

I appreciate any feedback and guidance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,6 +347,19 @@ make calls to retrieve account/budget information.  We recommend using the
   # now you can do all the normal api calls
   # ex:
   mint.get_transaction_data()
+  
+  # Update a transaction
+  # Note that the list of tag_ids replaces the existing tags on the transaction, rather than adding to them.
+  updated_transaction = mint.update_transaction_data(
+    transaction_id="15123456_2345678910_0",
+    description="McDonald's",
+    notes="Lunch",
+    category_id="15123456_1234",
+    tag_ids=["15123456_1234567"],
+    transaction_type="CashAndCreditTransaction",
+  )
+  
+
 ```
 
 ---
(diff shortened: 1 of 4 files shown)


burkematthew:
I'm still thinking on this one. We've so far maintained the stance that this is simply a read-only library. That being said, I do get that if this is a sought after feature and we have contributors willing to do the work, then that stance may need to be reconsidered. If we do,  I think we should work on major version v3 that supports a variety of update actions instead of slow trickling them in with various v2 minor versions.

mrooney:
Thanks, this looks cool! I don't have any strong opinions. As Matthew said, previously we've maintained a read-only stance, partially because I would be concerned about there being enough folks available and interested to maintain it. That said, it is probably easier with the REST endpoints.

I think my preferred approach would be a new repo under the same project umbrella named `mintapi-write` or similar that depends on mintapi and simply adds some endpoints (and tests) so it is more modularized. I imagine this might involve some tweaks to the mintapi project to make things easier to import and extend. If I made a separate empty repo, would you be willing to make it a new module @CalvinChanCan ?

CalvinChanCan:
Thanks @mrooney . I'm unsure if I have enough experience to create a new module. Specifically is the goal of the new module to try to create a wrapper around mintapi with write endpoints?

mrooney:
Apologies for the long delay @CalvinChanCan ! If you are still willing to work on this and a separate module seems overwhelming, I would be fine with just separating files, so that we have an `endpoints.py` and an `endpoints_write.py` or similar, and corresponding for the tests. This way, it keeps it all separate and we can import it from other files and makes it easier in the future to make a separate module. Does that make sense?

Issue #603, opened 2023-06-17
Getting TypeError: __init__() got an unexpected keyword argument 'executable_path'` when running python example


dloza:
`Traceback (most recent call last):
  File "/Users/dloza/mintupdater/tmp.py", line 3, in <module>
    mint = mintapi.Mint(
  File "/usr/local/lib/python3.9/site-packages/mintapi/api.py", line 33, in __init__
    self.browser = SeleniumBrowser(
  File "/usr/local/lib/python3.9/site-packages/mintapi/browser.py", line 57, in __init__
    self.login_and_get_token(
  File "/usr/local/lib/python3.9/site-packages/mintapi/browser.py", line 113, in login_and_get_token
    self.driver = driver or _create_web_driver_at_mint_com(
  File "/usr/local/lib/python3.9/site-packages/mintapi/signIn.py", line 313, in _create_web_driver_at_mint_com
    driver = Chrome(
  File "/usr/local/lib/python3.9/site-packages/seleniumrequests/request.py", line 148, in __init__
    super(RequestsSessionMixin, self).__init__(*args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'executable_path'`

yongjunkim42:
Selenium 4.10 has this known issue(https://github.com/SeleniumHQ/selenium/issues/12180).  You can temporarily fix this issue by rolling Selenium to 4.9.

basnijholt:
`pip install "selenium<4.10"`

adrienhenneron:
I've got the same error and rolling Selenium to 4.9 doesn't fix the issue

CalvinChanCan:
The issue seems to be that selenium fully depreciated `executable_path` in 4.10 (https://github.com/SeleniumHQ/selenium/issues/12176) but its still being used in here:

https://github.com/mintapi/mintapi/blob/main/mintapi/signIn.py#L315

jprouty:
A better workaround is to use the `use_chromedriver_on_path` arg (or kwarg for cli). If you have a recent selenium version, this will use selenium manager under the hood and is way better/the right way to go by default going forward.

@basnijholt workaround stops working, as the Chromium/ChromeDriver team has changed how releases work, so the old way mintapi uses to download and use the latest version of ChromeDriver stopped working with Chrome release 114. (See `get_stable_chrome_driver`)

jprouty:
[...]

PR #614, opened 2023-08-24
fix: update chromedriver download site and resolve selenium executable path deprecation

### LINKED ISSUE #610: Chromedriver for Chrome 116 issue
Hi,

I upgraded to Chrome 116 and now I am getting this error:
selenium.common.exceptions.SessionNotCreatedException: Message: session not created: This version of ChromeDriver only supports Chrome version 114
Current browser version is 116.0.5845.97 with binary path C:\Program Files\Google\Chrome\Application\chrome.exe

I downloaded the new chromedriver from this url https://googlechromelabs.github.io/chrome-for-testing/ but I still get the same error, has anyone else experienced this?

Thank you,

Paul

CalvinChanCan:
Resolves #610 and #603

Starting with chromedriver 115+, the chrome team changed the site where new chromedrivers are released. The old site only has chromedriver 114 available while the new site has simplified how to get the latest stable version and download link. 

Old: https://sites.google.com/chromium.org/driver/downloads

New: https://googlechromelabs.github.io/chrome-for-testing/

Determining latest stable version with download link:
https://googlechromelabs.github.io/chrome-for-testing/last-known-good-versions-with-downloads.json

I've tested this on Windows and Mac (arm) but have not tried on Mac x86 or Linux.

Edit: Also included a fix for #603. Starting with Selenium 4.10.0, they have now fully depreciated executable_path and replaced it with a Service object.


mintapi/signIn.py
```diff
@@ -283,11 +279,28 @@ def get_stable_chrome_driver(download_directory=os.getcwd()):
         )
 
     zip_file = zipfile.ZipFile(io.BytesIO(request.content))
-    zip_file.extractall(path=download_directory)
```
snowskeleton, line 286: What problem are we solving by rewriting this function?

snowskeleton (approved):
These changes worked to resolve my chromedriver download issues, on macOS and on Linux. All changes make sense, save the one comment I left. Very well written.

mintapi/signIn.py
CalvinChanCan: @snowskeleton Thanks for the feedback! 

The issue we're solving here is that the new chromedriver download zip now have a subdirectory (i.e. `chromedriver-win64`) instead of the chromedriver binary and license files. So this new function extracts the files from the subdirectory to the download directory whereas the `extractall` function just extracts the subdirectory to the download directory causing some filepath issues.

Old chromdriver download with no subdirectory:
https://chromedriver.storage.googleapis.com/114.0.5735.90/chromedriver_win32.zip

New chromedriver downoad with subdirectory:
https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/116.0.5845.96/win32/chromedriver-win32.zip

snowskeleton:
I just realized this patch isn't backwards compatible. Do we want to make < v115 users manually download their chromedrivers? That might be a fine option, with how fast things move. It might be nice to support both for a bit at least, maybe in such a way that the lower version code can be easily deleted.

CalvinChanCan:
@snowskeleton Good point. I was looking into one solution which is extremely simple but requires upgrading Selenium.

To fix #603, we have to use Selenium 4.10.0 due the `executable_path` deprecation. Though starting in Selenium 4.6.0, Selenium introduced a new [Selenium Manager](https://www.selenium.dev/blog/2022/introducing-selenium-manager/) that handles downloading the chromedriver binary for the matching version and platform. Though due to #610, Selenium Manager had issues with finding chromedrivers 115+ but was fixed in Selenium 4.11.2. 

This way, we can remove the logic to download the chromedriver binary and should allow us to still support chromedriver < v115. Though the downside is that we would have to set the minimum version of Selenium to 4.11.2. Thoughts?

Example:
```
def _create_web_driver_at_mint_com(
    headless=False,
    session_path=None,
    # use_chromedriver_on_path=False,
    # chromedriver_download_path=os.getcwd(),
):
    """
    Handles starting a web driver at mint.com
    """
    chrome_options = ChromeOptions()
    if headless:
        chrome_options.add_argument("headless")
        chrome_options.add_argument("no-sandbox")
        chrome_options.add_argument("disable-dev-shm-usage")
        chrome_options.add_argument("disable-gpu")
        # chrome_options.add_argument("--window-size=1920x1080")
    if session_path is not None:
        chrome_options.add_argument("user-data-dir=%s" % session_path)

    # if chromedriver is not in the project directory, selenium manager will download chromedriver in .cache/selenium. 
    return Chrome(options=chrome_options)
```

snowskeleton:
I'm not aware of a problem bumping Selenium to 4.11.2 with the thought process you laid out. It sounds like this is the happy path being laid out by the Selenium devs, which is a good place to be.

mrooney:
Thanks for the work! I'm definitely happy to see any improvements and modernization to this area, as it seems like a common pain / breaking point. LGTM!

mrooney:
@CalvinChanCan wow, that removes a lot of code, nice! Is t he Selenium Manager thing totally implicit and Just Works in 4.11.2+? I'm not seeing any code actually handles the downloading / etc.

mrooney (approved):
Thanks!

mrooney:
Thanks again, released in mintapi==2.15 :)

PR #621, opened 2023-10-12
fix: cli bills calling wrong endpoint

jaron-l:
### Summary
While using the CLI version, I noticed that the bills option is not mentioned in the README.md and that the pathway was broken. I added it to the README.md and fixed the function  call.

### Testing
I tested it locally by running the CLI with the bills argument and received data.


burkematthew (approved):

mrooney:
Awesome, thanks!

mrooney:
Thanks again, released in mintapi==2.15 :)

PR #624, opened 2023-11-26
Added a CLI hook for existing API method for account balances

Ryanman:
Hi folks. For anyone looking to leave mint, I think pulling balances out of accounts to get history is going to be key.

I am more of a .NET guy, so please forgive any errors and feel free to make any requests.

### DIFF
--- a/mintapi/endpoints.py
+++ b/mintapi/endpoints.py
@@ -820,11 +820,10 @@ class MintEndpoints(object, metaclass=ABCMeta):
         """
         # Account Balance history by account
         balances = []
-
         accounts = self.get_account_data()
         for account in accounts:
             account_id = account["id"]
-            # try extracting as asset and debt
+            # Try extracting as asset and debt
             for report in (
                 ReportView.Options.ASSETS_TIME,
                 ReportView.Options.DEBTS_TIME,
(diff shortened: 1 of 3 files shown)


mintapi/endpoints.py
burkematthew, line 1: @Ryanman can you explain why we need to manipulate the data? For the majority of the time we try to avoid doing such data manipulation and let this serve as a simple interface with Mint's endpoints. Couldn't your script do this sequence instead of relying on mintapi to do it?

Ryanman: Out of these changes, there's really two parts:
* Enriching the response with Account Name/Type from the accounts endpoint. I can see how this violates the ethos, but both properties are included in the balances UI of Mint, and would be obviously useful if you're trying to join balances with something. Which a lot of people are going to need to do in the next month or so!
* The isZillow flag, which is included in the balance API response already but is just straight up wrong - false for every account type.

Ryanman: @burkematthew just let me know if either this seems useful to merge in its entirety, without the additional helpful key columns, or if I should withdraw it!

burkematthew (changes requested):
@Ryanman given that Mint is pending a shutdown, I'll approve this no problem.  That being said, there is a black error to correct.

burkematthew (approved):

mrooney:
Awesome, thanks! If this helps folks migrate, that's great!

Ryanman:
> Awesome, thanks! If this helps folks migrate, that's great!

Thank you guys so much for building this, I so wish I had found it earlier. It's taking a little of the sting out of the shutdown!

